PR #22, opened 2020-11-10
Update `checkOverlappingPaths` function

### LINKED ISSUE #21: Bug: Paths don't allow using path names that starts with a string that is used in other path names
## what
* Bug: Paths don't allow using path names that starts with a string that is used in other path names

## why

We use the following configuration in `vendir.yml`:

```
  - path: components/terraform
    contents:
      - path: account-map
        git:
          url: https://github.com/cloudposse/terraform-aws-components
          ref: 0.135.0
        newRootPath: modules/account-map
        includePaths:
          - modules/account-map/**/*

      - path: account
        git:
          url: https://github.com/cloudposse/terraform-aws-components
          ref: 0.135.0
        newRootPath: modules/account
        includePaths:
          - modules/account/**/*
```

and getting the following error:

```
Error: Parsing resource config 'vendir.yml': Unmarshaling config: Validating config: Expected to not manage overlapping paths: 'components/terraform/account-map' and 'components/terraform/account'
```

Note that `account_map` does not work either, throwing the same validation error.

Moving the folders to the top-level throws the same error as well:

```
- path: components/terraform/account-map
    contents:
      - path: .
        git:
          url: https://github.com/cloudposse/terraform-aws-components
          ref: 0.135.0
        newRootPath: modules/account-map
        includePaths:
          - modules/account-map/**/*
  - path: components/terraform/account
    contents:
      - path: .
        git:
          url: https://github.com/cloudposse/terraform-aws-components
          ref: 0.135.0
        newRootPath: modules/account
        includePaths:
          - modules/account/**/*
```

This does not work either:

```
      - path: accountmap
        git:
          url: https://github.com/cloudposse/terraform-aws-components
          ref: 0.135.0
        newRootPath: modules/account-map
        includePaths:
          - modules/account-map/**/*

      - path: account
        git:
          url: https://github.com/cloudposse/terraform-aws-components
          ref: 0.135.0
        newRootPath: modules/account
        includePaths:
          - modules/account/**/*

```

throwing the error:

```
Error: Parsing resource config 'vendir.yml': Unmarshaling config: Validating config: Expected to not manage overlapping paths: 'components/terraform/accountmap' and 'components/terraform/account'

```

But this works ok:

```
      - path: a-ccountmap
        git:
          url: https://github.com/cloudposse/terraform-aws-components
          ref: 0.135.0
        newRootPath: modules/account-map
        includePaths:
          - modules/account-map/**/*

      - path: account
        git:
          url: https://github.com/cloudposse/terraform-aws-components
          ref: 0.135.0
        newRootPath: modules/account
        includePaths:
          - modules/account/**/*
```

So the issue is that the code checks not for the full path, but if a path starts with a string.

@cppforlife can you please take a look at the issue?
We like `vendir` and want to use it in our architectures.

Thank you

aknysh:
## what
* Update `checkOverlappingPaths` function

## why
* Fixes #21

### DIFF
--- a/pkg/vendir/config/config.go
+++ b/pkg/vendir/config/config.go
@@ -2,12 +2,10 @@ package config
 
 import (
 	"fmt"
-	"path/filepath"
-	"strings"
-
 	"github.com/ghodss/yaml"
 	semver "github.com/hashicorp/go-version"
 	"github.com/k14s/vendir/pkg/vendir/version"
+	"path/filepath"
 )
 
 const (
@@ -251,7 +249,7 @@ func (c Config) checkOverlappingPaths() error {
 
 	for i, path := range paths {
 		for i2, path2 := range paths {
-			if i != i2 && strings.Contains(path2, path) {
+			if i != i2 && path == path2 {
 				return fmt.Errorf("Expected to not "+
 					"manage overlapping paths: '%s' and '%s'", path2, path)
 			}
--- a/pkg/vendir/config/config.go
+++ b/pkg/vendir/config/config.go
@@ -2,10 +2,12 @@ package config
 
 import (
 	"fmt"
+	"path/filepath"
+	"strings"
+
 	"github.com/ghodss/yaml"
 	semver "github.com/hashicorp/go-version"
 	"github.com/k14s/vendir/pkg/vendir/version"
-	"path/filepath"
 )
 
 const (
@@ -249,7 +251,7 @@ func (c Config) checkOverlappingPaths() error {
 
 	for i, path := range paths {
 		for i2, path2 := range paths {
-			if i != i2 && path == path2 {
+			if i != i2 && strings.HasPrefix(path2+string(filepath.Separator), path+string(filepath.Separator)) {
 				return fmt.Errorf("Expected to not "+
 					"manage overlapping paths: '%s' and '%s'", path2, path)
 			}
--- a/.gitignore
+++ b/.gitignore
@@ -5,3 +5,5 @@
 /tmp
 .DS_Store
 /test/e2e/assets/simple-app/empty-dir
+.idea
+*.iml


pivotal-issuemaster:
@aknysh Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/pivotal?repositoryId=k14s/vendir&pullRequestId=22)!

[Click here](https://cla.pivotal.io/sync/pivotal?repositoryId=k14s/vendir&pullRequestId=22) to manually synchronize the status of this Pull Request.

See the [FAQ](https://cla.pivotal.io/about) for frequently asked questions.

pivotal-issuemaster:
@aknysh Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/pivotal?repositoryId=k14s/vendir&pullRequestId=22)!

pkg/vendir/config/config.go
```diff
@@ -251,7 +249,7 @@ func (c Config) checkOverlappingPaths() error {
 
 	for i, path := range paths {
 		for i2, path2 := range paths {
-			if i != i2 && strings.Contains(path2, path) {
+			if i != i2 && path == path2 {
```
cppforlife: this isnt going to catch case of /foo/bar/blah and /foo/bar overlapping. instead we do something like this i think `strings.HasPrefix(path2+filepath.Separator, path+filepath.Separator)`. this anchors to the beginning of the path and since it adds a separator at the end it will not have false positives like in the issue you filed.

pkg/vendir/config/config.go
```diff
@@ -2,12 +2,10 @@ package config
 
 import (
 	"fmt"
-	"path/filepath"
-	"strings"
-
 	"github.com/ghodss/yaml"
 	semver "github.com/hashicorp/go-version"
 	"github.com/k14s/vendir/pkg/vendir/version"
+	"path/filepath"
```
cppforlife: style convention: std lib imports go first with a newline separator for external libs.

aknysh: I just ran Go format code, but can revert it back

pkg/vendir/config/config.go
aknysh: yes I see it, will update

danielhelfand (commented):
The [existing test](https://github.com/k14s/vendir/blob/a1dd2b1fc373a2fbc3eb1968adb8400cafe1e185/test/e2e/overlapping_dir_err_test.go) in place might also need an update to account for the new approach.

aknysh:
> The [existing test](https://github.com/k14s/vendir/blob/a1dd2b1fc373a2fbc3eb1968adb8400cafe1e185/test/e2e/overlapping_dir_err_test.go) in place might also need an update to account for the new approach.

@danielhelfand  I looked at the existing test, they should work w/o changes (e.g. they will fail as expected using the new updates).
Do you have something else in mind for the test?

pkg/vendir/config/config.go
aknysh: updated

pkg/vendir/config/config.go
aknysh: updated

danielhelfand:
> @danielhelfand I looked at the existing test, they should work w/o changes (e.g. they will fail as expected using the new updates).
> Do you have something else in mind for the test?

My apologies, I reviewed before your most recent change. So yes, it should continue working now. Thanks for checking though!

aknysh:
@cppforlife @danielhelfand 
any chance you can review the PR?
thanks

cppforlife (approved):

aknysh:
@cppforlife @danielhelfand
can you please merge it?
(we are already using it in our repos, just waiting on the fix)

cppforlife:
will be included in the next release.

aknysh:
@cppforlife @danielhelfand
very sorry to bother you, but when do you think you would cut a new release?
we pin all our modules and components to releases (not using `master` anywhere), and we kind of waiting on the fix.
thank you for all your help.

PR #43, opened 2021-01-25
Add support for imgpkg bundles

### LINKED ISSUE #34: support imgpkg bundle download

want to download imgpkg bundles.

```yaml
imgpkgBundle:
  # image URL; could be plain, tagged or digest reference (required)
  image: gcr.io/repo/image:v1.0.0
  # specifies name of a secret with registry auth details;
  # secret may include 'username', 'password' and/or 'token' keys (optional)
  secretRef:
    # (required)
    name: my-image-auth
```

would use imgpkg pull -b ...

TBD is is too confusing that we are using `image` for referring to bundle? mostly interested in doing that because kbld picks up `image` key automatically for image ref.

related: should we update image source to use image key instead of url.

alexbrand:
I've tested this with a custom build of kapp-controller.

This is still missing an example to go in the `examples/` directory but otherwise works.

Please let me know if I have missed anything else.

Fixes #34

### DIFF
--- a/pkg/vendir/config/directory.go
+++ b/pkg/vendir/config/directory.go
@@ -32,6 +32,7 @@ type DirectoryContents struct {
 	Git           *DirectoryContentsGit           `json:"git,omitempty"`
 	HTTP          *DirectoryContentsHTTP          `json:"http,omitempty"`
 	Image         *DirectoryContentsImage         `json:"image,omitempty"`
+	Bundle        *DirectoryContentsBundle        `json:"bundle,omitempty"`
 	GithubRelease *DirectoryContentsGithubRelease `json:"githubRelease,omitempty"`
 	HelmChart     *DirectoryContentsHelmChart     `json:"helmChart,omitempty"`
 	Manual        *DirectoryContentsManual        `json:"manual,omitempty"`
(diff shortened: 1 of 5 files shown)


pkg/vendir/fetch/imgpkgbundle/sync.go
```diff
@@ -0,0 +1,98 @@
...
+}
+
+func NewSync(opts ctlconf.DirectoryContentsImgpkgBundle, refFetcher ctlfetch.RefFetcher) *Sync {
+	return &Sync{opts, refFetcher}
+}
+
+var (
+	// Example image ref in imgpkg stdout:
+	//   Pulling bundle 'index.docker.io/dkalinin/consul-helm@sha256:d1cdbd46561a144332f0744302d45f27583fc0d75002cba473d840f46630c9f7'
+	imgpkgPulledImageRef = regexp.MustCompile("(?m)^Pulling bundle '(.+)'$")
```
cppforlife, line 27: @joaopapereira would be good to figure out a better way to do this instead of parsing imgpkg output. we do have --lock-ouptut but can we somehow grab it without temp file?

cppforlife: (not a blocker for this PR imho)

pkg/vendir/config/directory.go
```diff
@@ -293,6 +309,17 @@ func (c *DirectoryContentsImage) Lock(lockConfig *LockDirectoryContentsImage) er
 	return nil
 }
 
+func (c *DirectoryContentsImgpkgBundle) Lock(lockConfig *LockDirectoryContentsImgpkgBundle) error {
+	if lockConfig == nil {
+		return fmt.Errorf("Expected image lock configuration to be non-empty")
+	}
+	if len(lockConfig.Image) == 0 {
+		return fmt.Errorf("Expected image Image to be non-empty")
```
ewrenn8: Should this be "Expected bundle Image to be non-empty"?

alexbrand: Addresed. PTAL. Thanks!

PR #44, opened 2021-01-26
When a release is published check that github action works

joaopapereira:


.github/workflows/release-published.yml
```diff
@@ -0,0 +1,25 @@
...
+    steps:
+      - uses: k14s/setup-k14s-action@v1
+        with:
+          only: vendir
+      - run: |
+          vendir version
+          version=`vendir version`
+          tag="${{ github.event.release.tag_name }}"
+          tool_version="$(echo $version | sed 's/^.*[^0-9]\([0-9]*\.[0-9]*\.[0-9]*\).*$/\1/')"
+          if [[ "v${tool_version}" == "${tag}" ]];
```
jbrunton, line 18: There's an edge case where this check may not do what you expect: if you provide a minor or patch update to a previous major version (e.g. for security reasons). The setup action is designed to install the latest version by semver name by default, not the most recent version published to GitHub if that is different.

Admittedly, this is very much an edge case. However, if you want to anticipate it, I _think_ you could amend the setup step like this:

```yml
- uses: k14s/setup-k14s-action@v1
  with:
    only: vendir
    vendir: ${{ github.event.release.tag_name }}
```

Note that I didn't test this, but in theory this should install the correct version for the release tag, even if you've just published a patch for an older major/minor version.

EDIT: indeed, if your only requirement is to verify the checksum, then the second step (to compare the tool_version with the tag) wouldn't be necessary after this change. However, if you want to verify the published binary version matches the published tag, then that would of course still be necessary.

neil-hickey:
hey @joaopapereira I think this was done as part of https://github.com/vmware-tanzu/carvel-vendir/pull/133 right? Happy to close this out?

joaopapereira:
yet, going to close

PR #52, opened 2021-03-06
Feature/sync with tanzu network

crhntr:
## Example

```yaml
kind: Config
directories:
  - path: vendor
    contents:
      - path: ops-manager
        tanzuNetwork:
          slug: ops-manager
          version: ">= 2.7, < 2.8"
          files:
            - name: ops-manager-aws-*.yml
      - path: tas-cli
        tanzuNetwork:
          slug: elastic-runtime
          version: ">= 2.7, < 2.8"
          files:
            - name: cf-cli-*.zip
```
## Use case

This would allow the Platform Provider Experience Team (including TAS RelEng) to use vendir more.

This could also be used by Tanzu Support folks; however, I have not done any user research on this use case.

## TODO
- [x] make linter happy
- [ ] add end to end test.
    I am not sure how you would do this. Some secret TanzuNet api token would need to be added in CI or something.

## Questions
- Do we need to implement some way to declare acceptance of the EULA in the resource definition?
- Does this add too much to the lock file? I added the sha, file name, file id, and release version. I think this makes the lock file easier to reason about but only the file id is required to fetch an expected resource. The additional fields just make it easier to reason about the lockfile.


cppforlife:
@crhntr is there a CLI for tanzu network that can do downloading to a directory? one concern i have within vendir is putting in bunch of fetchers and getting stuck in go module dependency hell. this is one of the reasons we tried to maintain loose coupling to other CLIs: e.g. imgpkg, git, helm.

crhntr:
There is a [CLI for Tanzu Network](https://github.com/pivotal-cf/pivnet-cli). I could refactor this fetcher to use that or copy the bits of code from the pivnet module to make the few required API requests to the sever. Both solutions would reduce external dependencies. As a user I would prefer using the second solution as it required less set up but either would work.

cppforlife:
> copy the bits of code from the pivnet module to make the few required API requests to the sever

if you can make that work without adding any non-std library dependencies then im ok with that.

im also ok with calling out to tanzu network CLI. ill let you decide which way to go.

cppforlife:
closing due to inactivity. feel free to reopen.

PR #55, opened 2021-03-19
Add arm64 binary build

benjvi:
Allows vendir to be run on ARM64 linux machines, as you can already with kapp

I'm using vendir in a Raspberry Pi Github runner, haven't encountered any arch-specific problems with this so far (as expected).


danielhelfand (approved):
lgtm

Thanks @benjvi.

cc @ewrenn8 @cppforlife

PR #60, opened 2021-04-19
Helm OCI chart fetch

cppforlife:
this PR includes 3 commits:
- extract http source in helmchart => no change to functionality, just extraction. prep for next commit
- support experimental oci helm charts via oci:// repository url => supports repository.url oci:// schema. uses experimental commands in helm cli. it seems that eventually some of these commands would be merged into other commands.
- support dockerconfigjson secrets for image/imgpkgBundle/helmChart => semi related change to support dockerconfigjson secrets for specifying registry auth

this PR did not add e2e test for oci artifact fetching since it turns out most registries dont support custom artifact types. i could see eventually adding e2e tests against local registry:2, but that seems like a separable issue. meanwhile tested manually against a private registry i had access to.

### DIFF
--- a/pkg/vendir/config/dockerconfigjson.go
+++ b/pkg/vendir/config/dockerconfigjson.go
@@ -1,0 +1,52 @@
+// Copyright 2021 VMware, Inc.
+// SPDX-License-Identifier: Apache-2.0
+
+package config
+
+import (
+	"encoding/json"
+	"fmt"
+)
+
+type secretDockerConfigJSON struct {
+	Auths map[string]secretDockerConfigJSONAuth
+}
+
+type secretDockerConfigJSONAuth struct {
+	Username string
+	Password string
+}
+
+func (s Secret) ToBasicAuthSecret() (Secret, error) {
+	const (
+		// Constants from Kubernetes core v1
+		typeDockerConfigJSON = "kubernetes.io/dockerconfigjson"
+		dockerConfigJSONKey  = ".dockerconfigjson"
+	)
+
+	if s.Type != typeDockerConfigJSON {
+		return s, nil // return itself
+	}
+
+	var data secretDockerConfigJSON
+
+	err := json.Unmarshal(s.Data[dockerConfigJSONKey], &data)
+	if err != nil {
+		return Secret{}, err
+	}
+
+	if len(data.Auths) != 1 {
+		return Secret{}, fmt.Errorf("Expected exactly one registry configuration within a secret")
+	}
+
+	for _, auth := range data.Auths {
+		return Secret{
+			Data: map[string][]byte{
+				SecretK8sCorev1BasicAuthUsernameKey: []byte(auth.Username),
+				SecretK8sCorev1BasicAuthPasswordKey: []byte(auth.Password),
+			},
+		}, nil
+	}
+
+	panic("Unreachable")
+}
--- a/pkg/vendir/config/dockerconfigjson_test.go
+++ b/pkg/vendir/config/dockerconfigjson_test.go
@@ -1,0 +1,68 @@
+// Copyright 2021 VMware, Inc.
+// SPDX-License-Identifier: Apache-2.0
+
+package config_test
+
+import (
+	"reflect"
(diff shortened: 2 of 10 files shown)


pkg/vendir/fetch/helmchart/http_source.go
```diff
@@ -0,0 +1,166 @@
+// Copyright 2020 VMware, Inc.
```
danielhelfand: Nit: 2021 for year

pkg/vendir/fetch/helmchart/sync.go
```diff
@@ -63,21 +61,16 @@ func (t *Sync) Sync(dstPath string, tempArea ctlfetch.TempArea) (ctlconf.LockDir
...
+	if t.opts.Repository != nil && strings.HasPrefix(t.opts.Repository.URL, "oci://") {
+		err = NewOCISource(t.opts, t.helmBinary, t.refFetcher).Fetch(chartsDir, tempArea)
+		if err != nil {
+			return lockConf, err
+		}
+	} else {
+		err = NewHTTPSource(t.opts, t.helmBinary, t.refFetcher).Fetch(chartsDir, tempArea)
+		if err != nil {
+			return lockConf, err
+		}
```
danielhelfand, line 73: Could move `if err != nil` check out side if-else blocks to avoid duplication, but might make sense to keep for readability.

pkg/vendir/fetch/helmchart/oci_source.go
```diff
@@ -0,0 +1,198 @@
+// Copyright 2020 VMware, Inc.
```
danielhelfand: Nit: 2021 for year

pkg/vendir/fetch/helmchart/http_source.go
```diff
@@ -0,0 +1,166 @@
...
+
+	err := cmd.Run()
+	if err != nil {
+		return fmt.Errorf("Fetching helm chart: %s (stderr: %s)", err, stderrBs.String())
+	}
+
+	return nil
+}
+
+func (t *HTTPSource) addAuthArgs(args []string) ([]string, error) {
```
danielhelfand, line 144: Maybe a TODO here could be to have `HTTPSource`/`OCISource` implement an interface and then have add auth Args accept both as a param? But fine for now.

pkg/vendir/config/dockerconfigjson_test.go
```diff
@@ -0,0 +1,67 @@
+// Copyright 2021 VMware, Inc.
+// SPDX-License-Identifier: Apache-2.0
+package config_test
```
danielhelfand, line 4: Add space between comment and package

pkg/vendir/config/dockerconfigjson.go
```diff
@@ -0,0 +1,51 @@
+// Copyright 2021 VMware, Inc.
+// SPDX-License-Identifier: Apache-2.0
+package config
```
danielhelfand, line 4: Add space between comment and package

pkg/vendir/fetch/helmchart/sync.go
cppforlife: readability is what i would favor in this case. this duplication isnt a business related logic duplication so wouldnt lead to complications.

pkg/vendir/config/dockerconfigjson.go
cppforlife: fixed.

pkg/vendir/fetch/helmchart/oci_source.go
cppforlife: fixed.

pkg/vendir/fetch/helmchart/http_source.go
cppforlife: fixed.

pkg/vendir/config/dockerconfigjson_test.go
cppforlife: fixed.

PR #65, opened 2021-05-21
Honor (https?|no)_proxy environment variables for helmChart

jbertozzi:
Hello,

The helmChart sync does not honor (https?|no)_proxy environment variables making it impossible to use behind a proxy.

This PR fixes it.

Feel free to comment the code (I don't have a developer background) and I'll do my best to improve it.

Thanks to dkalinin for the help in slack.

Jérémy

### DIFF
--- a/pkg/vendir/fetch/helmchart/http_source.go
+++ b/pkg/vendir/fetch/helmchart/http_source.go
@@ -44,7 +44,7 @@ func (t *HTTPSource) Fetch(dstPath string, tempArea ctlfetch.TempArea) error {
 }
 
 func setEnv(helmHomeDir string) []string {
-	re := regexp.MustCompile(`(?i)^(https?|no)_proxy=.*`)
+	re := regexp.MustCompile(`(?i)\A(https?|no)_proxy=.*`)
 	env := []string{"HOME=" + helmHomeDir}
 	for _, e := range os.Environ() {
 		if re.MatchString(e) {
--- a/pkg/vendir/fetch/helmchart/http_source.go
+++ b/pkg/vendir/fetch/helmchart/http_source.go
@@ -8,6 +8,7 @@ import (
 	"fmt"
 	"os"
 	"os/exec"
+	"regexp"
 	"strings"
 
 	ctlconf "github.com/vmware-tanzu/carvel-vendir/pkg/vendir/config"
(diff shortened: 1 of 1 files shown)


pkg/vendir/fetch/helmchart/http_source.go
```diff
@@ -42,13 +43,24 @@ func (t *HTTPSource) Fetch(dstPath string, tempArea ctlfetch.TempArea) error {
 	return t.fetch(helmHomeDir, dstPath)
 }
 
+func SetEnv(helmHomeDir string) []string {
+	re := regexp.MustCompile(`(https?|no)_proxy=.*`)
```
cppforlife: let's also check for uppercase variants as well (e.g. HTTP_PROXY). regexp should also start with `\A` to anchor match to the beginning of string.

pkg/vendir/fetch/helmchart/http_source.go
```diff
@@ -42,13 +43,24 @@ func (t *HTTPSource) Fetch(dstPath string, tempArea ctlfetch.TempArea) error {
 	return t.fetch(helmHomeDir, dstPath)
 }
 
+func SetEnv(helmHomeDir string) []string {
```
cppforlife: let's turn this function private.

pkg/vendir/fetch/helmchart/http_source.go
jbertozzi: done

pkg/vendir/fetch/helmchart/http_source.go
jbertozzi: done

jbertozzi:
Thanks for the review, I have updated accordingly.

pkg/vendir/fetch/helmchart/http_source.go
```diff
@@ -42,13 +43,24 @@ func (t *HTTPSource) Fetch(dstPath string, tempArea ctlfetch.TempArea) error {
 	return t.fetch(helmHomeDir, dstPath)
 }
 
+func setEnv(helmHomeDir string) []string {
+	re := regexp.MustCompile(`(?i)^(https?|no)_proxy=.*`)
```
cppforlife: i think we should use `\A` instead of `^` since we may accidentally match middle of the string after a newline.

jbertozzi: Hi,

It has been fixed by this [commit](https://github.com/vmware-tanzu/carvel-vendir/pull/65/commits/de33a3d91927326a3f4bbf80ec2768f011cc9006).

Issue #68, opened 2021-05-28
vendir cannot access the file because it is being used by another process on windows


meier-christoph:
**What steps did you take:**

When you add more than 1 path to directories using the http plugin you get these errors on windows. 

```yaml
---
apiVersion: vendir.k14s.io/v1alpha1
kind: Config
directories:
  - path: vendor/one
    contents:
      - path: one
        http:
          url: https://raw.githubusercontent.com/vmware-tanzu/carvel-vendir/develop/examples/http/vendir.yml

  - path: vendor/two
    contents:
      - path: two
        http:
          url: https://raw.githubusercontent.com/vmware-tanzu/carvel-vendir/develop/examples/http/vendir.yml
```

I suspect that the tmp dir is not unique and/or is shared with both downloads that run in some goroutines in parallel but then they try to delete the dir (or parent dir).

**What happened:**

```
Error: Syncing directory 'vendor/two': Deleting tmp dir '.vendir-tmp': remove .vendir-tmp\incoming\vendir-http521295291: The process cannot access the file because it is being used by another process.
```

**Environment:**

- vendir version (execute `vendir --version`): v0.19.0
- OS (e.g. from `/etc/os-release`): Windows 10

---
Vote on this request

This is an invitation to the community to vote on issues, to help us prioritize our backlog. Use the "smiley face" up to the right of this comment to vote.

👍 "I would like to see this addressed as soon as possible"
👎 "There are other more important things to focus on right now"

We are also happy to receive and review Pull Requests if you want to help working on this issue.

cppforlife:
@meier-christoph did you have two "vendir sync" running in parallel? can you confirm that running a single "vendir sync" on your system is resulting with this error?

cppforlife:
the actual problem looks to be that we are not closing temp file [0]. even though we are trying to remove it we still need to do tmpFile.Close beforehand. fixing this issue should fix the clean up issue.

[0] https://github.com/vmware-tanzu/carvel-vendir/blob/05250dd89612f3fe07d9d5667e03ba70117a1f5a/pkg/vendir/fetch/http/sync.go#L34-L44

meier-christoph:
only a single `vendir sync`, nice find on the missing close, updated the pr #70

meier-christoph:
Fixed by #70

PR #70, opened 2021-06-04
fix: multiple issues when using vendir on windows

### LINKED ISSUE #67: vendir downloads helm charts to c:\windows
**What steps did you take:**

I am using the carvel suite with helm on windows. Any helm chart should be able to reproduce the issue.

```yaml
kind: Config
directories:
  - path: vendor
    contents:
      - path: kube-prometheus
        helmChart:
          name: kube-prometheus
          version: 5.0.0
          repository:
            url: https://charts.bitnami.com/bitnami
```

**What happened:**

```
Error: Syncing directory 'vendor': Syncing directory 'kube-prometheus' with helm chart contents: Fetching helm chart: exit status 1 (stderr: Error: failed to untar: mkdir C:\WINDOWS\helm-741980851: Access is denied.)
```

**What did you expect:**

It should use the `.vendir-tmp` dir relative to the current dir or use `~/.cache/vendir` or event `/tmp` or `C:\Users\{user}\AppData\Local\Temp`

**Environment:**

- vendir version (execute `vendir --version`): v0.19.0
- OS (e.g. from `/etc/os-release`): Windows 10

---
Vote on this request

meier-christoph:
fix: home dir on windows
fix: helm env vars
fix: conflict on tmp folders
fix: tmp folder not being deleted

Fixes #67 and #68

### DIFF
--- a/pkg/vendir/directory/staging_dir.go
+++ b/pkg/vendir/directory/staging_dir.go
@@ -7,7 +7,6 @@ import (
 	"fmt"
 	"io"
 	"io/ioutil"
-	"math/rand"
 	"os"
 	"path/filepath"
 	"strings"
@@ -23,7 +22,7 @@ type StagingDir struct {
 }
 
 func NewStagingDir() StagingDir {
-	rootDir := fmt.Sprintf(".vendir-tmp-%d", rand.Uint32())
+	rootDir := ".vendir-tmp"
 	return StagingDir{
 		rootDir:     rootDir,
 		stagingDir:  filepath.Join(rootDir, "staging"),
@@ -159,19 +158,7 @@ func (d StagingDir) TempArea() StagingTempArea {
 }
 
 func (d StagingDir) CleanUp() error {
-	// workaround where os.RemoveAll would not delete folder
-	var err error
-	for {
-		stat, e := os.Stat(d.rootDir)
-		if e == nil && stat.IsDir() {
-			err = d.cleanUpAll()
-		} else if !os.IsNotExist(e) {
-			return e
-		} else {
-			break
-		}
-	}
-	return err
+	return d.cleanUpAll()
 }
 
 func (d StagingDir) cleanUpAll() error {
@@ -191,9 +178,10 @@ var _ ctlfetch.TempArea = StagingTempArea{}
 func (d StagingTempArea) NewTempDir(name string) (string, error) {
 	dir := d.path
 	if dir == "" {
+		// same behavior as ioutil.TempFile()
 		dir = os.TempDir()
 	}
-	tmpDir := filepath.Join(dir, fmt.Sprintf("%s-%d", name, rand.Uint32()))
+	tmpDir := filepath.Join(dir, name)
 
 	absTmpDir, err := filepath.Abs(tmpDir)
 	if err != nil {
--- a/pkg/vendir/directory/staging_dir.go
+++ b/pkg/vendir/directory/staging_dir.go
@@ -176,12 +176,7 @@ type StagingTempArea struct {
 var _ ctlfetch.TempArea = StagingTempArea{}
 
 func (d StagingTempArea) NewTempDir(name string) (string, error) {
-	dir := d.path
-	if dir == "" {
-		// same behavior as ioutil.TempFile()
-		dir = os.TempDir()
-	}
-	tmpDir := filepath.Join(dir, name)
+	tmpDir := filepath.Join(d.path, name)
 
 	absTmpDir, err := filepath.Abs(tmpDir)
 	if err != nil {
--- a/pkg/vendir/directory/staging_dir.go
+++ b/pkg/vendir/directory/staging_dir.go
@@ -165,6 +165,8 @@ func (d StagingDir) CleanUp() error {
 		stat, e := os.Stat(d.rootDir)
 		if e == nil && stat.IsDir() {
 			err = d.cleanUpAll()
+		} else if !os.IsNotExist(e) {
+			return e
 		} else {
 			break
 		}
--- a/pkg/vendir/directory/staging_dir.go
+++ b/pkg/vendir/directory/staging_dir.go
@@ -7,6 +7,7 @@ import (
 	"fmt"
 	"io"
 	"io/ioutil"
+	"math/rand"
 	"os"
 	"path/filepath"
 	"strings"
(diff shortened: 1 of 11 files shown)


pkg/vendir/directory/staging_dir.go
```diff
@@ -85,7 +86,17 @@ func (d StagingDir) TempArea() StagingTempArea {
 }
 
 func (d StagingDir) CleanUp() error {
-	return d.cleanUpAll()
+	// workaround where os.RemoveAll would not delete folder
+	var err error
+	for {
+		stat, e := os.Stat(d.rootDir)
```
ewrenn8: The case exists where the stat fails for a reason other than a NotExist error. In this case, I would expect the method to return an error since cleanup failed, but in this implementation, it will return `nil`.

Is there any more info about the underlying reason that os.RemoveAll wasn't cleaning up the directory? If it's possible, I think getting to the bottom of that is more desirable than implementing a retry.

meier-christoph: I don't know what is causing the folder not being removed, I did try a few things prior to adding this hack.

- I disabled the AV as this would be the first suspect
- I added some delay at the end of the program to see if maybe the rm was done in the background and the prog exited to fast

the folders were still not deleted.

I will add a check on the error type as you suggested.

meier-christoph: I did some more testing and the os.RemoveAll actually fails with an error `The process cannot access the file because it is being used by another process.` but the error messages are droped because all the cleanup methods are used with `defer`

pkg/vendir/fetch/helmchart/http_source.go
```diff
@@ -45,12 +45,20 @@ func (t *HTTPSource) Fetch(dstPath string, tempArea ctlfetch.TempArea) error {
 
 func setEnv(helmHomeDir string) []string {
 	re := regexp.MustCompile(`(?i)\A(https?|no)_proxy=.*`)
-	env := []string{"HOME=" + helmHomeDir}
+	env := []string{
+		"HOME=" + helmHomeDir,
+		"TEMP=" + helmHomeDir,
+		"HELM_CACHE_HOME=" + helmHomeDir,
+		"HELM_CONFIG_HOME=" + helmHomeDir,
+		"HELM_DATA_HOME=" + helmHomeDir,
```
cppforlife, line 53: is helm not deriving these based on $HOME?

pkg/vendir/directory/staging_dir.go
```diff
@@ -19,7 +20,7 @@ type StagingDir struct {
 }
 
 func NewStagingDir() StagingDir {
-	rootDir := ".vendir-tmp"
+	rootDir := fmt.Sprintf(".vendir-tmp-%d", rand.Uint32())
```
cppforlife: per this comment: https://github.com/vmware-tanzu/carvel-vendir/issues/68#issuecomment-856792616, i dont believe we need this change.

pkg/vendir/directory/staging_dir.go
```diff
@@ -103,7 +116,11 @@ type StagingTempArea struct {
 var _ ctlfetch.TempArea = StagingTempArea{}
 
 func (d StagingTempArea) NewTempDir(name string) (string, error) {
-	tmpDir := filepath.Join(d.path, name)
+	dir := d.path
+	if dir == "" {
+		dir = os.TempDir()
+	}
+	tmpDir := filepath.Join(dir, fmt.Sprintf("%s-%d", name, rand.Uint32()))
```
cppforlife: we want to contain all our temporary work under staging dir (.vendir-tmp).

pkg/vendir/fetch/helmchart/http_source.go
cppforlife: does this change alone fix the error you were receiving?

pkg/vendir/directory/staging_dir.go
meier-christoph: removed the rand, keept the os.TempDir to be consistent with ioutil.TempFile

pkg/vendir/fetch/helmchart/http_source.go
meier-christoph: on windows APPDATA and TEMP are also required, those 3 take precedence on all systems and as such we have everything inside .vendir-tmp

meier-christoph: https://helm.sh/docs/helm/helm/

cppforlife: if you are saying APPDATA and TEMP are required, should we add those, and remove HELM_*?

pkg/vendir/directory/staging_dir.go
```diff
@@ -103,7 +103,12 @@ type StagingTempArea struct {
 var _ ctlfetch.TempArea = StagingTempArea{}
 
 func (d StagingTempArea) NewTempDir(name string) (string, error) {
-	tmpDir := filepath.Join(d.path, name)
+	dir := d.path
+	if dir == "" {
+		// same behavior as ioutil.TempFile()
+		dir = os.TempDir()
+	}
+	tmpDir := filepath.Join(dir, name)
```
cppforlife: we want to use staging dir as temp directory instead of OS specific location because we want to track all files we create in one place. given that there is no need to add this code (especially that we fixed tmpFile.Close() issue).

meier-christoph: yeah I know, its just that I don't agree. 

Having this check doesn't do harm and if someone forgets to set the path by accident in the future, then maybe it could prevent an other issue like #67.

This kind of defensive programming is used all over in go (and other languages) and as I already said you are using methods in this project that fallback to os.TempDir() already.

Anyway, this kind of stuff is not worth arguing over so I reverted the change.

pkg/vendir/fetch/helmchart/http_source.go
meier-christoph: It should work as well but then you should probably add APPDATA only on windows and you would have to change the logic for each OS (more code and more stuff to test).

The HELM_* take precedence anyway and are compatible with all OS, I think this is better.

aaronshurley:
@cppforlife @ewrenn8 @danielhelfand could one of y'all take a look at this?

cppforlife:
looks like test suite isnt happy about some configuration in vendor/modules.txt, @meier-christoph would you mind trying to run ./hack/build.sh with go1.16+

meier-christoph:
I did a rebase on the latest changes, please try again.

ewrenn8:
We will try and cut a release soon to get these updates out.  Thanks for the contribution @meier-christoph!

PR #79, opened 2021-06-24
Add fallback to sorting build meta for determinism

ewrenn8:


ewrenn8:
Constraints don't work in this PR and aren't easily supported on top of the library code. Closing in favor of forking the library and implementing there.

Issue #84, opened 2021-07-07
💡 Implement scoop 💡 


adriens:
# Context

[Scoop](https://scoop.sh/) is a very popular windows distribution system, in the same philosophy of brew, but for windows. The cosystem is growing very fast, so the idea would to : 

- [x] Create a first local running install of `vendir`
- [ ] Do it more widely through a central repo (like brew tap), which is called a [Scoop Bucket](https://github.com/lukesampson/scoop/wiki/Buckets)

**NB : with a dedicated `Carvel` bucket, it would be very nice to install all toher carvel tools**

# Usage

Please take a look at [SCOOP_GUIDELINES.md](SCOOP_GUIDELINES.md) for how to use it, but it's as easy as : 

```
git clone https://github.com/vmware-tanzu/carvel-vendir.git
cd carvel-vendir
export TARGET_VERSION=0.21.1
sed 's/$VENDIR_VERSION/'"$TARGET_VERSION"'/g' vendir.json.template > vendir.json
scoop install vendir
```

# Additional resources

- [A nice intro](https://ibigbug.online/how-to-publish-a-scoop-package)

adriens:
Pull request under moderation on Scoop main repo : 

https://github.com/ScoopInstaller/Main/pull/2411

adriens:
Badge : 

```
![Scoop Version](https://img.shields.io/scoop/v/vendir)
```

![Scoop Version](https://img.shields.io/scoop/v/vendir)

PR #85, opened 2021-07-07
Implement Scoop install.

adriens:
First scoop attempt see #84

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,3 +8,4 @@
 /test/e2e/assets/simple-app/empty-dir
 .idea
 *.iml
+/vendir.json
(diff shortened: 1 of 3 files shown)


cppforlife:
cc @DennisDenuto since you are exploring goreleaser integrations.

DennisDenuto:
@adriens I'm not super knowledgable around the windows ecosystem wrt package managers... but...

Can we use `scoop` instead of `chocolatey` to distribute our binary files to windows folk?

`goreleaser` *does* support generating `scoop` files and would 'plug-in' smoothly to some of our other tools release process which i'm trying to automate (some of the work is still in draft).

for e.g.

if we were to use goreleaser to generate the scoop files, we could do something similar to this: https://github.com/vmware-tanzu/carvel-imgpkg/pull/180 (except replace the word homebrew with scoop)

Issue #10, opened 2020-08-11
Feature Request: Add semver support


ciriarte:
## Context

**vendir** currently supports tracking specific refs and versions. I'd like to have support for specific version ranges:

```diff
  apiVersion: vendir.k14s.io/v1alpha1
  kind: Config
  directories:
  - path: vendor
    contents:
    - path: github.com/cloudfoundry/cf-k8s-networking
      git:
        url: https://github.com/cloudfoundry/cf-k8s-networking
-       ref: master
+       version: ^1.0.0
    includePaths:
    - cfroutesync/crds/**/*
    - install/ytt/networking/**/*
  - path: github.com/cloudfoundry/cc
    manual: {}
  - path: github.com/GoogleCloudPlatform/metacontroller
    manual: {}
  - path: local-dir
    directory:
      path: local-dir
```

So I can specify valid version ranges for each one of my dependencies.

cppforlife:
i wonder if we should make this functionality more generic since vendir currently supports git and github release, but will also support sources such as docker images, helm charts, etc. ... or does it make sense to just keep it source specific?

paulcwarren:
+1 for this.  

The syntax I have used for this in the past was something like this BTW `[1.0.0, 2.0.0)`; i.e. get the latest available from 1.0.0 up to but excluding 2.0.0.

ciriarte:
> i wonder if we should make this functionality more generic since vendir currently supports git and github release, but will also support sources such as docker images, helm charts, etc. ... or does it make sense to just keep it source specific?

Yes. It shouldn't be specific to a source.

ram-pivot:
+1 It will really helpful as we are also planning to consume other external components as libraries and this would help us to pin to a particular version.

cppforlife:
sources:

- [ ] git (via tags)
- [ ] githubRelease (via tags)
- [ ] helm chart (via versions)
- [ ] image (via tags)
- [ ] http (???)

anEXPer:
The Pivnet resource (and various other Concourse-ecosystem resources) accomplish some of this with regex filters; this allows expressive constraints of what versions will be included/excluded without having to create/document/maintain yet-another version specification DSL, or adopt one of the various (sometimes pretty complex) existing ones.

This also allows a distinction between the "explicit string" `version` key and a mutually-exclusive `version-regex` string, if desired.

Why not allow, for example,
`version-regex: ^v1\..*`

[...]

PR #90, opened 2021-07-29
Semver for Github Release Tags

benjvi:
This was mentioned as one of the sources to apply semver to in #10 , and is a gap I've been running into in my use of vendir

Some choices I made that seem worth calling out:
- Had to split up `Desc()` and `URL()` calls to avoid the tag list being fetched twice. I kept running into Github rate limits so this seems like a necessary optimisation
- I factored out the common API calling code from `downloadRelease` into a separate function that can be called by `downloadTags` and `downloadReleases` 
- I ignored pagination to keep this initial change simple. It seems the max number of results per page is 100 so you would most likely only run into this on a larger project when requesting older versions


pkg/vendir/fetch/githubrelease/sync.go
```diff
@@ -7,6 +7,7 @@ import (
 	"crypto/sha256"
 	"encoding/json"
 	"fmt"
+	ctlver "github.com/vmware-tanzu/carvel-vendir/pkg/vendir/versions/v1alpha1"
```
cppforlife, line 15: nit: can you move this import into section below where non-std imports are

pkg/vendir/fetch/githubrelease/sync.go
```diff
@@ -165,17 +190,71 @@ func (d Sync) matchesAssetName(name string) (bool, error) {
...
+		tags = append(tags, tag.Name)
+	}
+	tag, err := ctlver.HighestConstrainedVersion(tags, *d.opts.TagSelection)
+	if err != nil {
+		return "", fmt.Errorf("Failed to find tag matching tagSelection %v : %s", d.opts.TagSelection.Semver, err)
+	}
+	return tag, err
+}
+
+func (d Sync) downloadTags(authToken string) ([]TagAPI, error) {
```
cppforlife: im a bit afraid adding this feature in without pagination since it may provide unexpected results based on what gets published latest (general problem with tag selection in other places as well eg gotta fetch all tags in git).

pkg/vendir/fetch/githubrelease/sync.go
```diff
@@ -165,17 +190,71 @@ func (d Sync) matchesAssetName(name string) (bool, error) {
...
 
-	_, url, err := d.DescAndURL()
+	url, err := d.URL()
+	if err != nil {
+		return releaseAPI, fmt.Errorf("getting release URL: %s", err)
+	}
+	respBytes, err := d.downloadAPIResponse(url, authToken)
 	if err != nil {
-		return releaseAPI, err
+		errMsg := err.Error()
```
cppforlife: nit: inline errMsg

pkg/vendir/fetch/githubrelease/sync.go
```diff
@@ -165,17 +190,71 @@ func (d Sync) matchesAssetName(name string) (bool, error) {
...
+		return releaseAPI, fmt.Errorf("error downloading release details from %s : %s", url, errMsg)
 	}
 
+	err = json.Unmarshal(respBytes, &releaseAPI)
+	if err != nil {
+		return releaseAPI, fmt.Errorf("error parsing response from: %s error: %s", url, err.Error())
```
cppforlife: nit: we use capitalized error messages. also dont mention "error" since it's already part of error.

pkg/vendir/fetch/githubrelease/sync.go
```diff
@@ -30,24 +31,48 @@ func NewSync(opts ctlconf.DirectoryContentsGithubRelease,
...
+		}
 	case d.opts.Latest:
 		desc = d.opts.Slug + "@latest"
+	default:
+		return "", fmt.Errorf("Expected to have non-empty tag, tagSelection, latest or url")
+	}
+	return desc, nil
+}
+
+func (d Sync) URL() (string, error) {
```
cppforlife: lets turn this into private method

cppforlife (commented):
left a few comments. thanks for kicking this off.

benjvi:
@cppforlife I think this is ready for a second pass. Could you have a look? After looking into pagination for a short while, I came across the `go-github` client library which seems like a better approach than writing that sort of API-calling code again here in vendir, so I made the changes to the list tags methods to use that instead

pkg/vendir/fetch/githubrelease/sync.go
```diff
@@ -165,17 +210,70 @@ func (d Sync) matchesAssetName(name string) (bool, error) {
 	return false, nil
 }
 
+func (d Sync) fetchTagSelection() (string, error) {
+	opt := &github.RepositoryListByOrgOptions{
+		ListOptions: github.ListOptions{PerPage: 10},
```
cppforlife: might as well crank it up if we are going to be fetching all of them. how about 50?

pkg/vendir/fetch/githubrelease/sync.go
```diff
@@ -165,17 +210,70 @@ func (d Sync) matchesAssetName(name string) (bool, error) {
...
+	}
+
+	tags := []string{}
+	for {
+		tagList, resp, err := d.client.Repositories.ListTags(context.Background(), strings.Split(d.opts.Slug, "/")[0], strings.Split(d.opts.Slug, "/")[1], &github.ListOptions{})
```
cppforlife: lets pull out `strings.Split(d.opts.Slug, "/")[0]` and `strings.Split(d.opts.Slug, "/")[1]` as vars outside of the loop with names like ownerName, repoName.

pkg/vendir/fetch/githubrelease/sync.go
```diff
@@ -165,17 +210,70 @@ func (d Sync) matchesAssetName(name string) (bool, error) {
 	return false, nil
 }
 
+func (d Sync) fetchTagSelection() (string, error) {
+	opt := &github.RepositoryListByOrgOptions{
```
cppforlife: who should be using this? ListTags?

cppforlife: can we modify tagSelection test to pick up tag from non-0th page?

benjvi: Whoops yeah, should be passed to list tags. 

I *think* it was , will need to revisit and also make it more explicit though, especially if we'll increase the page size used

pkg/vendir/fetch/githubrelease/sync.go
benjvi: Sounds sensible to me

benjvi: I adjusted this to 40 so I can continue to use the same repo for the test case. That repo happens to have exactly 50 tags right now, so this way we can fetch from the second page

PR #94, opened 2021-08-13
Hg improve

cppforlife:


evankanderson:
Hey, it looks like this PR committed a set of AGPL source code as an example. This sets of license scanners, because not everyone is thrilled with the viral nature of AGPL.

Could we replace the code in `examples/hg/vendor/exact-sha` with some other project that's not AGPL licensed?

joaopapereira:
@evankanderson do you mind opening an issue with this? thanks

PR #95, opened 2021-08-13
Image auth

cppforlife:
addresses https://github.com/vmware-tanzu/carvel-vendir/issues/93


pkg/vendir/fetch/helmchart/oci_source.go
```diff
@@ -176,21 +176,28 @@ func (t *OCISource) addAuthArgs(args []string) ([]string, io.Reader, error) {
...
-			case ctlconf.SecretK8sCorev1BasicAuthPasswordKey:
-				authArgs = append(authArgs, []string{"--password-stdin"}...)
-				passwordStdin = strings.NewReader(string(val))
-			default:
-				return nil, nil, fmt.Errorf("Unknown secret field '%s' in secret '%s'", name, secret.Metadata.Name)
+		for _, secret := range secrets {
+			for name, val := range secret.Data {
+				switch name {
+				case ctlconf.SecretRegistryHostnameKey:
+					// do nothing for now
```
joe-kimmel-vmw, line 188: should there be a log message / warning that we're skipping this case?

pkg/vendir/fetch/helmchart/oci_source.go
```diff
@@ -176,21 +176,28 @@ func (t *OCISource) addAuthArgs(args []string) ([]string, io.Reader, error) {
...
+				case ctlconf.SecretK8sCorev1BasicAuthUsernameKey:
+					authArgs = append(authArgs, []string{"--username", string(val)}...)
+				case ctlconf.SecretK8sCorev1BasicAuthPasswordKey:
+					authArgs = append(authArgs, []string{"--password-stdin"}...)
+					passwordStdin = strings.NewReader(string(val))
+				default:
+					return nil, nil, fmt.Errorf("Unknown secret field '%s' in secret '%s'", name, secret.Metadata.Name)
+				}
 			}
+			// Only single set of credentials is supported
```
joe-kimmel-vmw, line 199: this is a little opaque -- is the behavior that we're looping through the secrets until we find the first one that looks like it has auth credentials?   might be good to advertise that somehow,  whether in docstrings, user-facing documentation -  not sure who needs to know,  but seems like a potential stumbling block for a naive operator.

joe-kimmel-vmw (commented):
i feel underqualified to comment on the imgpkg use which is honestly probably the stickier bit. mostly just want to make sure that nobody will be surprised by the behaviors w.r.t. credentials.

pkg/vendir/fetch/helmchart/oci_source.go
cppforlife: 👍 docs update coming to carvel repo.

pkg/vendir/fetch/helmchart/oci_source.go
cppforlife: dont think it would be that useful to the user to see this. will make a note in docs though.

pkg/vendir/fetch/helmchart/oci_source.go
cppforlife: simplified logic for now to only allow 0/1 secrets: https://github.com/vmware-tanzu/carvel-vendir/commit/bdb4ba833bbb40f34d3c70635d8ae516059d9f56

PR #107, opened 2021-09-16
support extracting username+password from auth field if password is empty

cppforlife:


pkg/vendir/config/dockerconfigjson.go
```diff
@@ -52,6 +56,20 @@ func (s Secret) ToRegistryAuthSecrets() ([]Secret, error) {
...
 
+		if len(auth.Password) == 0 && len(auth.Auth) > 0 {
+			decodedAuth, err := base64.StdEncoding.DecodeString(auth.Auth)
+			if err != nil {
+				return nil, fmt.Errorf("Decoding auth field: %s", err)
+			}
+
+			pieces := strings.SplitN(string(decodedAuth), ":", 2)
+			if len(pieces) != 2 {
+				return nil, fmt.Errorf("Expected auth field to have username+password, but did not")
```
danielhelfand:
```suggestion
				return nil, fmt.Errorf("Expected auth field to have username:password format, but did not")
```

PR #110, opened 2021-09-23
[hg] set custom configuration options

WnP:
At the moment we cannot use [hg evolve](https://www.mercurial-scm.org/doc/evolution/) with the current version of `vendir` because it require the following lines in hgrc file:

```
[extensions]
evolve =
topic =
```
This pull request add a new option to hg directories named `hgrc` so we can do like following:

```yaml
kind: Config
directories:
- path: somewhere
  contents:
  - path: myrepo
    hg:
      url: https://somewhere.io/my/hg/repo
      ref: my-topic
      hgrc: |
        [extensions]
        evolve =
        topic =
      secretRef:
        name: my-auth
```

I'm not sure if:
- this option should be global?
- we should only make a special case for this one and only add an option like `evolve: true`?

Let me know your thoughts.

### DIFF
--- a/pkg/vendir/fetch/hg/hg.go
+++ b/pkg/vendir/fetch/hg/hg.go
@@ -88,7 +88,11 @@ func (t *Hg) fetch(dstPath string, tempArea ctlfetch.TempArea) error {
 		return err
 	}
 
-	var authRc string
+	var hgRc string
+
+	if len(t.opts.Hgrc) > 0 {
+		hgRc = fmt.Sprintf("%s", t.opts.Hgrc)
+	}
 
 	if authOpts.Username != nil && authOpts.Password != nil {
 		if !strings.HasPrefix(hgURL, "https://") {
@@ -99,11 +103,12 @@ func (t *Hg) fetch(dstPath string, tempArea ctlfetch.TempArea) error {
 			return fmt.Errorf("Parsing hg remote url: %s", err)
 		}
 
-		authRc = fmt.Sprintf(`[auth]
+		hgRc = fmt.Sprintf(`%s
+[auth]
 hgauth.prefix = https://%s
 hgauth.username = %s
 hgauth.password = %s
-`, hgCredsURL.Host, *authOpts.Username, *authOpts.Password)
+`, hgRc, hgCredsURL.Host, *authOpts.Username, *authOpts.Password)
 
 	}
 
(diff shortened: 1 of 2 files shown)


pkg/vendir/fetch/hg/hg.go
```diff
@@ -99,11 +103,12 @@ func (t *Hg) fetch(dstPath string, tempArea ctlfetch.TempArea) error {
 			return fmt.Errorf("Parsing hg remote url: %s", err)
 		}
 
-		authRc = fmt.Sprintf(`[auth]
+		hgRc = fmt.Sprintf(`%s
```
cppforlife, line 106: i think we should do hgRc += (and may be add a newline before) instead of overriding it entirely.

pkg/vendir/fetch/hg/hg.go
```diff
@@ -134,25 +139,25 @@ hgauth.password = %s
...
 		}
 
-		authRc = fmt.Sprintf("%s\n[ui]\nssh = %s\n", authRc, strings.Join(sshCmd, " "))
+		hgRc = fmt.Sprintf("%s\n[ui]\nssh = %s\n", hgRc, strings.Join(sshCmd, " "))
 	}
 
-	if (authOpts.Username != nil && authOpts.Password != nil) || authOpts.IsPresent() {
-		credsHgRcPath := filepath.Join(authDir, "hgrc")
-		err = ioutil.WriteFile(credsHgRcPath, []byte(authRc), 0600)
+	if (authOpts.Username != nil && authOpts.Password != nil) || authOpts.IsPresent() || len(t.opts.Hgrc) > 0 {
```
cppforlife: this could be just simplified if hgRc is non-empty.

cppforlife:
top level comment:

> At the moment we cannot use hg evolve with the current version of vendir because it require the following lines in hgrc file:

after reading a bit more about mercurial evolve i dont quite understand why it needs to be in the temporary hgrc for cloning. it seems that evolve is more about workflows when working with mercurial, which is not something that vendir needs during cloning.

are you getting some error without these options?

pkg/vendir/fetch/hg/hg.go
WnP: In fact I need to string format for other arguments, so overwriting it in this case use only one `fmt.Sprintf` call, otherwise it require one `fmt.Sprintf` call and then a `+=` call.

Let me know if you still want that.

pkg/vendir/fetch/hg/hg.go
WnP: fixed :heavy_check_mark:

WnP:
@cppforlife, evolve enables mercurial [topic](https://blog.orus.io/post/mercurial-evolve-and-topics/) feature, which is something like a git branch. During development I sometime use to refer a chart revision which is on a given topic, at the moment if I try to checkout onto a topic using its name or its changeset (SHA1) vendir (well mercurial) return the following error:

```sh
Error: Syncing directory 'vendor/ytt': Syncing directory 'myrepo' with hg contents: Fetching hg repository: Hg [checkout mytopic]: exit status 255 (stderr: abort: unknown revision 'mytopic')
# same apply using SHA1
Error: Syncing directory 'vendor/ytt': Syncing directory 'myrepo' with hg contents: Fetching hg repository: Hg [checkout 71f3be55056908e55a148800d1d540cfcd38c2bb]: exit status 255 (stderr: abort: unknown revision '71f3be55056908e55a148800d1d540cfcd38c2bb')
```
Adding extensions settings to hgrc fix the issue.

cppforlife:
hey @WnP, ive been thinking a bit more on this...

> we should only make a special case for this one and only add an option like evolve: true?

lets roll with this option for now (separate PR?). im a little concerned opening up all of hgrc which i imagine may include some configuration that would not be perhaps safe to blindly consume from untrusted sources.

cppforlife:
@WnP bump ^. any thoughts?

WnP:
Hi @cppforlife,

Sorry for the delay I was quite busy.

I've created #119 with the required changes. Let me know your thoughts.

cppforlife:
closing in favor of https://github.com/vmware-tanzu/carvel-vendir/pull/119 merged.

PR #134, opened 2022-02-15
Use sigs.k8s.io/yaml instead of ghodss/yaml

benmoss:
I don't think there's a need for both, standardizing on one

xref https://github.com/vmware-tanzu/carvel-kapp-controller/pull/520


neil-hickey (approved):
LGTM

PR #138, opened 2022-03-03
Error on invalid symlinks

benmoss:
- Defaults to enabled, opt out with --dangerous-allow-all-symlink-destinations
- Requires that symlinks resolve to files inside the directory being synced
- Requires that symlinks resolve to actual files

### DIFF
--- a/pkg/vendir/cmd/sync.go
+++ b/pkg/vendir/cmd/sync.go
@@ -5,6 +5,7 @@ package cmd
 
 import (
 	"fmt"
+	"io/fs"
 	"os"
 	"path/filepath"
 	"strings"
@@ -30,7 +31,8 @@ type SyncOptions struct {
 	Directories []string
 	Locked      bool
 
-	Chdir string
+	Chdir                       string
+	AllowAllSymlinkDestinations bool
 }
 
 func NewSyncOptions(ui ui.UI) *SyncOptions {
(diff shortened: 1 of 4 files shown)


pkg/vendir/cmd/sync.go
```diff
@@ -50,6 +52,7 @@ func NewSyncCmd(o *SyncOptions) *cobra.Command {
 	cmd.Flags().BoolVarP(&o.Locked, "locked", "l", false, "Consult lock file to pull exact references (e.g. use git sha instead of branch name)")
 
 	cmd.Flags().StringVar(&o.Chdir, "chdir", "", "Set current directory for process")
+	cmd.Flags().BoolVarP(&o.AllowAllSymlinkDestinations, "dangerous-allow-all-symlink-destinations", "", false, "Symlinks to all destinations are allowed")
```
cppforlife: nit: no need to use P since we dont have a single-letter shorthand.

pkg/vendir/cmd/sync.go
```diff
@@ -126,6 +129,31 @@ func (o *SyncOptions) Run() error {
...
+				if info.Type()&os.ModeSymlink == os.ModeSymlink {
+					resolvedPath, err := filepath.EvalSymlinks(path)
+					if err != nil {
+						return fmt.Errorf("Unable to resolve symlink: %w", err)
+					}
+					absPath, err := filepath.Abs(resolvedPath)
+					if err != nil {
+						return err
+					}
+					if !strings.HasPrefix(absPath, absRoot) {
```
cppforlife: should we be paranoid and check that it's, either

absPath == absRoot, or
HasPrefix(absPath, absRoot+string(filepath.Separator))

otherwise we could allow pointing to destinations that are "sibling" to allowed dir.

(i did similar logic here: https://github.com/vmware-tanzu/carvel-kapp-controller/blob/develop/pkg/memdir/scoped.go)

pkg/vendir/cmd/sync.go
```diff
@@ -126,6 +129,31 @@ func (o *SyncOptions) Run() error {
 		if err != nil {
 			return fmt.Errorf("Syncing directory '%s': %s", dirConf.Path, err)
 		}
+		absRoot, err := filepath.Abs(dirConf.Path)
```
cppforlife: - can move this into if conditional
- also would be nice to call it absRoot => allowedAbsRoot

PR #141, opened 2022-03-10
dont set IMGPKG_ANON when no auth is explicitly given

cppforlife:
removal of this allow imgpkg to use other auth methods; this is inline with allowing git fetcher to use its regular auth mechanisms even though they are not explicitly called out in vendir.yml

### DIFF
--- a/pkg/vendir/fetch/image/imgpkg.go
+++ b/pkg/vendir/fetch/image/imgpkg.go
@@ -120,10 +120,6 @@ func (t *Imgpkg) authEnv() ([]string, error) {
 		}
 	}
 
-	if len(authEnv) == 0 {
-		authEnv = []string{"IMGPKG_ANON=true"}
-	}
-
 	return authEnv, nil
 }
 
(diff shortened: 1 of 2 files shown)


davidreuss:
Works like a charm! ⭐ 

```
❯ ~/Code/vmware-tanzu/carvel-vendir/vendir sync
Fetching: vendor + dreuss/imgpkg-demo (imgpkgBundle from gcr.io/$gcr_private_project/imgpkg-demo:test)

Lock config

directories:
- contents:
  - imgpkgBundle:
      image: gcr.io/$gcr_private_project/imgpkg-demo@sha256:a05cd4917f9b48b57f872feeaeee2adffe746713bffad5ce4fa0e5c55e964b35
      tag: test
    path: imgpkg-demo
  path: vendor
kind: LockConfig

Succeeded
```

[Slack thread](https://kubernetes.slack.com/archives/CH8KCCKA5/p1646912970151439) on carvel channel in kubernetes slack workspace for reference.

Issue #145, opened 2022-03-30
githubRelease options should support downloading the source archive


voor:
**Describe the problem/challenge you have**
Currently github releases also include the source archive for the given tag.  Examples:
https://github.com/vmware-tanzu/carvel-vendir/archive/refs/tags/v0.26.0.tar.gz
https://github.com/vmware-tanzu/carvel-vendir/archive/refs/tags/v0.26.0.tar.gz

These are listed as `Assets` on the GitHub GUI, but are unable to be added to `assetNames` in the `githubRelease`.

**Describe the solution you'd like**
Support something that can allow referencing the source code archives in githubRelease that takes into consideration the resulting filename is dynamically generated off the tag name.  That means if somebody wanted to do something like this:
```
apiVersion: vendir.k14s.io/v1alpha1
kind: Config
minimumRequiredVersion: 0.12.0
directories:
  # Use this if you're downloading manifests directly from a location.
  - path: bundle/config/upstream
    contents:
      - path: .
        githubRelease:
          slug: aws/amazon-eks-pod-identity-webhook
          tagSelection:
            semver:
              constraints: ">=0.4.0"
          disableAutoChecksumValidation: true
          assetNames: ["*.tar.gz"]
          unpackArchive:
            path: "*.tar.gz"
        includePaths:
          - deploy/
        newRootPath: deploy/
```
They would need to refer to the assetNames and unpackArchive with wildcards, or some other kind of variable name to know when it's the source archive.

**Anything else you would like to add:**
This is currently remediated by just using `git` to pull down the given tag.

---
Vote on this request

This is an invitation to the community to vote on issues, to help us prioritize our backlog. Use the "smiley face" up to the right of this comment to vote.

👍 "I would like to see this addressed as soon as possible"
👎 "There are other more important things to focus on right now"

We are also happy to receive and review Pull Requests if you want to help working on this issue.

joe-kimmel-vmw:
Thanks @voor --  it seems like we'd have to introduce new flexibility around [how we build the url](https://github.com/vmware-tanzu/carvel-vendir/blob/develop/pkg/vendir/fetch/githubrelease/sync.go#L77).  

@cppforlife  is this feature in scope for vendir, given that there's a workaround of "just use git" ?

PR #152, opened 2022-04-13
Added support for `http` fetch within `githubRelease`

niallthomson:
This change allows the optional use of an `http` section within a `githubRelease` block, which can be used to directly create the URL of the asset to download with metadata from the GitHub release such as the tag name. The `url` parameter of `http` can interpolate the tag of the GitHub release using the `{tag}` token.

This is useful for several use-cases, primarily when a project publishes GitHub releases but does not include files as assets directly. It will also provide a solution for #145.

Example usage:

```
kind: Config
directories:
- path: vendor
  contents:
  # templates http url with release tag
  - path: github.com/kubernetes/kubectl
    githubRelease:
      slug: kubernetes/kubernetes
      tagSelection:
        semver:
          constraints: "~v1.23.x"
      http:
        url: https://dl.k8s.io/release/{tag}/bin/linux/amd64/kubectl
```

### DIFF
--- a/examples/github-release/vendir.lock.yml
+++ b/examples/github-release/vendir.lock.yml
@@ -13,5 +13,8 @@ directories:
   - githubRelease:
       url: https://api.github.com/repos/cloudfoundry/eirini-release/releases/23064766
     path: github.com/cloudfoundry-incubator/eirini-release
+  - githubRelease:
+      url: https://api.github.com/repos/kubernetes/kubernetes/releases/54824547
+    path: github.com/kubernetes/kubectl
   path: vendor
 kind: LockConfig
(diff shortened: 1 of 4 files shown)


niallthomson:
Looks like the tests are failing due to API request limits or am I missing something?

cppforlife:
@niallthomson would you mind rebasing on develop. i pushed a fix for symlink failures we've seen (seems like github action error detecting logic somehow changed and starting seeing it as error).

it does seem like there is a rate limit problem if it doesnt "disappear" we can configure to use GITHUB_TOKEN.

niallthomson:
@cppforlife rebased and pushed

cppforlife:
test failure is unrelated. im going to merge and resolve the test failure separately by setting github token in the CI. thanks for contributing @niallthomson.

PR #168, opened 2022-06-28
[hg] fault tolerant commit sha resolution

WnP:
if `terminfo` is wrongly configured `hg id -i --debug` can return the following errors:

```
no terminfo entry for invis
no terminfo entry for blink
no terminfo entry for invis
no terminfo entry for blink
no terminfo entry for invis
```
Which causes vendir to fail to retrieve the commit sha.

Using `hg log -r . -T '{node}'` returns the same result and solves this issue.

### DIFF
--- a/pkg/vendir/fetch/hg/hg.go
+++ b/pkg/vendir/fetch/hg/hg.go
@@ -48,8 +48,8 @@ func (t *Hg) Retrieve(dstPath string, tempArea ctlfetch.TempArea) (HgInfo, error
 
 	info := HgInfo{}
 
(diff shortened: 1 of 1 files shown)


cppforlife:
will try to cut a release shortly (seeing if we can merge another open PR before that)

cppforlife:
@WnP in https://github.com/vmware-tanzu/carvel-vendir/releases/tag/v0.28.0

WnP:
Thanks a lot @cppforlife

cppforlife:
ill share release glory with @DennisDenuto :)

PR #173, opened 2022-07-14
add depth to git fetcher

cppforlife:


DennisDenuto:
LGTM

- Ran the tests locally
- Manual test works with a commit ref
- If a given depth doesn't contain the commit ref the error message is a bit cryptic *but* it is consistent with providing a commit ref that doesn't exist. 
```
Running 'vendir sync'...
    vendir.go:79: Failed to successfully execute 'vendir sync': Execution error: stdout: '' stderr: 'vendir: Error: Syncing directory 'vendor':
          Syncing directory 'deeper-sha' with git contents:
            Fetching git repository:
              Git [-c advice.detachedHead=false checkout 5d06e4025f13e1230be4df632b915a4233e0694d]: exit status 128 (stderr: fatal: reference is not a tree: 5d06e4025f13e1230be4df632b915a4233e0694d
        )
```

@cppforlife If you haven't already, I think a follow up PR to the carvel docs should be made.

PR #178, opened 2022-08-22
Enable caching imgpkg

### LINKED ISSUE #160: Cache image and bundle contents
Implementation of https://github.com/vmware-tanzu/carvel-kapp-controller/issues/664

Allow the users to cache the contents of an image or bundle

Depending on the path forward defined in https://github.com/vmware-tanzu/carvel-imgpkg/issues/390 vendir will need to change the way it invokes imgpkg.

In terms of UX for vendir will now look for an environment variable `VENDIR_CACHE_DIR` and if it is present it will check for the presence of a cache result.
When caching is present it will just copy from the cache folder to the destination.

Use Cases:
---
**Use case 1:** Copied from cache when sync happens twice
 
**Given** I have the following configuration
```yaml
kind: Config
directories:
- path: vendor
  contents:
  - path: docker.io/dkalinin/consul-helm-by-digest
    image:
      url: index.docker.io/dkalinin/consul-helm@sha256:d1cdbd46561a144332f0744302d45f27583fc0d75002cba473d840f46630c9f7
    excludePaths:
    - .git/**/*
```
**When** I execute `VENDIR_CACHE_DIR=/tmp/vendir-cache vendir sync`
**Then** It should download the contents of the image
**Then** when I clear the current folder
**Then** I execute `VENDIR_CACHE_DIR=/tmp/vendir-cache vendir sync`
**And** vendir will populate the current folder with the contents that are present in `/tmp/vendir-cache/sha256-d1cdbd46561a144332f0744302d45f27583fc0d75002cba473d840f46630c9f7`

**Use case 2:** When image is not cacheable it does not save any data on the provided folder
 
**Given** I have the following configuration
```yaml
kind: Config
directories:
- path: vendor
  contents:
  - path: docker.io/dkalinin/consul-helm-by-tag
    image:
      url: dkalinin/consul-helm:latest
    excludePaths:
    - .git/**/*  
```
**When** I execute `VENDIR_CACHE_DIR=/tmp/vendir-cache vendir sync`
**Then** It should download the contents of the image
**And** the folder `/tmp/vendir-cache` should be empty

**Use case 3:** When image is not cacheable and max-cacheable-size is small than the content it **does NOT** fail and nothing is cached
 
**Given** I have the following configuration
```yaml
kind: Config
directories:
- path: vendor
  contents:
  - path: docker.io/dkalinin/consul-helm-by-tag
    image:
      url: dkalinin/consul-helm:latest
    excludePaths:
    - .git/**/*  
```
**When** I execute `VENDIR_CACHE_DIR=/tmp/vendir-cache VENDIR_MAX_CACHE_SIZE=1 vendir sync`
**Then** It should download the contents of the image
**And** The folder `/tmp/vendir-cache` is empty

Depending on the implementation provided by imgpkg, vendir might need to ensure that it is backward compatible with the version of imgpkg installed in the system.

vendir should use the SHA of the image as the key to store and retrieve the contents from the folder

The default value for the Maximum size of the folder to be cached is "1Mi"

Caching folder structure:
Format expected: `.vendir-cache/fetcher/artifactType/folderWithID`
- `artifactType` is the type of artifact that is being saved in the cache. Currently, we only have `image` and `imgpkgBundle` types with this feature. 
- `folderWithID` is a normalized version of the SHA of the images being cached. In the future this definition will evolve once we introduce caching for other type of artifacts being downloaded by vendir.

---
Vote on this request

We are also happy to receive and review Pull Requests if you want to help work on this issue.

joaopapereira:
This PR includes the following changes:
- Uses richgo, when present, to run tests to have an easier-to-read UX
- Start using imgpkg as a library
  - The only API available at this point is Pull so we still need the binary installed when the tags command is called
  - Included temporary hack that needs to be removed when vmware-tanzu/carvel-imgpkg#397 is fixed
- Allow the users to specify caching folder and the maximum size of caching they want to enable using the environment variables `VENDIR_CACHE_DIR` and `VENDIR_MAX_CACHE_SIZE` respectively.

 
Fixes #160


hack/patch-k8s-pkg-credentialprovider.patch
```diff
@@ -0,0 +1,13 @@
+diff --git a/vendor/github.com/vdemeester/k8s-pkg-credentialprovider/provider.go b/vendor/github.com/vdemeester/k8s-pkg-credentialprovider/provider.go
+index 8c9ad34..f953bb4 100644
+--- a/vendor/github.com/vdemeester/k8s-pkg-credentialprovider/provider.go
++++ b/vendor/github.com/vdemeester/k8s-pkg-credentialprovider/provider.go
+@@ -70,7 +70,7 @@ type CachingDockerConfigProvider struct {
+ 
+ // Enabled implements dockerConfigProvider
+ func (d *defaultDockerConfigProvider) Enabled() bool {
+-	return true
++	return false
```
joe-kimmel-vmw: this is like forking an open source project but sneakier.   is this a long-term plan?

pkg/vendir/fetch/cache/cache.go
```diff
@@ -0,0 +1,173 @@
...
+	q, err := resources.ParseQuantity(maxContentCacheableSize)
+	if err != nil {
+		return nil, fmt.Errorf("Unable to process maximum amount allowed to cache: %s", err)
+	}
+	return &FolderCache{folder: cacheFolder, maxSize: q}, nil
+}
+
+// Hit checks if a particular entry in the cache is present
+// Returns the path to the entry and a flag information if the entry was found or not
+func (c FolderCache) Hit(id string) (string, bool) {
```
joe-kimmel-vmw: nit: but to my mind "Hit" sounds more like a command than a query. of course "Present" has the same issues and is a longer word.  idk, for 3 letter words, ... "Has" ?

pkg/vendir/fetch/cache/resources/amount.go
```diff
@@ -0,0 +1,302 @@
+// Copyright 2022 VMware, Inc.
```
joe-kimmel-vmw, line 1: what does it mean to stick our copyright above something that has a different copyright?

pkg/vendir/fetch/image/imgpkg.go
```diff
@@ -135,3 +238,29 @@ func (t *Imgpkg) addDangerousArgs(args []string) []string {
 	}
 	return args
 }
+
+// Logger provided to the imgpkg API calls
+type Logger struct {
```
joe-kimmel-vmw, line 256: ok i'll admit,  i'm confused why, in a PR about caching, there's a logger that looks like it has levels but actually they're hardcoded and you'll never know which level something was logged at, especially if it was a debug log bc those don't even log at all? 

My guess is you needed an interface to pass to something else, but maybe some more elaborate hint in the comments would be helpful

joe-kimmel-vmw: i guess in general --  i like log messages.  like being able to get debug output.  i have some dreams w.r.t . being able to trigger all the tools in kapp-controller into debug at the same time, for the sake of figuring out what's going on.

joe-kimmel-vmw (commented):
i have left you a sort of incomplete code review but i think that's all i can give you at the moment.  thanks for letting me take a look.

hack/patch-k8s-pkg-credentialprovider.patch
joaopapereira: we tried to give provide a fix to the upstream repository but they told us that they wanted to keep the repository as a copy-paste from Kubernetes authentication libraries.
This is a temporary solution until we fix the issue https://github.com/vmware-tanzu/carvel-imgpkg/issues/397

pkg/vendir/fetch/cache/resources/amount.go
joaopapereira: we do add our copyright to all the files with cod in the project and we have been told that this is what we should be doing

pkg/vendir/fetch/image/imgpkg.go
joaopapereira: The logger created in this particular instance is a bit peculiar because it needs to conform to the requirements from the imgpkg API.
Because of that, it is only present in the `image` package. In the future when we are using our tools and can provide directives to activate different logging levels we should definitely revisit all these instances to ensure that we are doing the correct thing.
In the meanwhile, I will update the logger struct comment to inform the user that not all the logs will be present

pkg/vendir/fetch/cache/cache.go
joaopapereira: I was trying to use the caching more common terms like hit/miss to provide describe the behavior, but I think this can be an easy change

cppforlife:
i just realized that by merging this, it would make vendir block in environments where ambient creds endpoints are present but blocked (same problem that imgpkg has). this is especially nasty if user is not even using imgpkg fetcher (previously this was not a problem because imgpkg's dependencies are not executed when vendir itself starts). id like us to see ambient cred blocking problem fixed before we release vendir to avoid this backwards compat breaking behaviour.

joaopapereira:
@cppforlife what do you mean by ambient credentials blocking problem?

joaopapereira:
Are you talking about selectively choose the auth keychains?
From this issue https://github.com/vmware-tanzu/carvel-imgpkg/issues/326
Also the fix that is coming on this PR https://github.com/vmware-tanzu/carvel-imgpkg/pull/409 ?

cppforlife:
> Are you talking about selectively choose the auth keychains?

yup, im getting a little lost with one of the PRs closed and the other merged. looks like it the feature that we needed got merged. it seems that those changes were introduced in imgpkg v0.32.0. im ok merging this PR as is assuming we make another PR with bump to imgpkg v0.32.0 before we cut a next vendir release. (or feel free to revise this PR to do the bump).

pkg/vendir/fetch/cache/cache.go
```diff
@@ -0,0 +1,175 @@
...
+			return err
+		}
+		defer srcFile.Close()
+		_, err = io.Copy(f, srcFile)
+		return err
+	})
+}
+
+func (c FolderCache) idToFolder(artifactType string, id string) string {
+	normalizedID := strings.ReplaceAll(id, ":", "-")
```
cppforlife: should we turn id into a base64 or sha to avoid having to "try to sanitize" it?

cppforlife (commented):
left one comment before merge

PR #187, opened 2022-09-28
Avoid writing the lock config file if there are no changes

petewall:
When using `vendir sync -l`, if the lock file did not change, then do not write the lock file.

This preserves the original modified time, which makes it work much better with tools like Make.

Signed-off-by: Pete Wall <pwall@vmware.com>

### DIFF
--- a/pkg/vendir/cmd/sync.go
+++ b/pkg/vendir/cmd/sync.go
@@ -97,9 +97,8 @@ func (o *SyncOptions) Run() error {
 
 	// If syncing against a lock file, apply lock information
 	// on top of existing config
-	var existingLockConfig ctlconf.LockConfig
 	if o.Locked {
-		existingLockConfig, err = ctlconf.NewLockConfigFromFile(o.LockFile)
+		existingLockConfig, err := ctlconf.NewLockConfigFromFile(o.LockFile)
 		if err != nil {
 			return err
 		}
@@ -177,9 +176,6 @@ func (o *SyncOptions) Run() error {
 		return nil
 	}
 
-	if newLockConfig.IsEqualTo(existingLockConfig) {
-		return nil
-	}
 	return newLockConfig.WriteToFile(o.LockFile)
 }
 
--- a/pkg/vendir/cmd/sync.go
+++ b/pkg/vendir/cmd/sync.go
@@ -97,8 +97,9 @@ func (o *SyncOptions) Run() error {
 
 	// If syncing against a lock file, apply lock information
 	// on top of existing config
+	var existingLockConfig ctlconf.LockConfig
 	if o.Locked {
-		existingLockConfig, err := ctlconf.NewLockConfigFromFile(o.LockFile)
+		existingLockConfig, err = ctlconf.NewLockConfigFromFile(o.LockFile)
 		if err != nil {
 			return err
 		}
@@ -176,6 +177,9 @@ func (o *SyncOptions) Run() error {
 		return nil
 	}
 
+	if newLockConfig.IsEqualTo(existingLockConfig) {
+		return nil
+	}
 	return newLockConfig.WriteToFile(o.LockFile)
 }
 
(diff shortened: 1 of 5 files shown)


pkg/vendir/cmd/sync.go
```diff
@@ -176,6 +177,9 @@ func (o *SyncOptions) Run() error {
 		return nil
 	}
 
+	if newLockConfig.IsEqualTo(existingLockConfig) {
```
joaopapereira: Couldn't this if statement be inside the `o.Locked` if statement?
Because if it is not locked the existingLockConfig will be empty struct which is not equal to the newLockConfig for sure

pkg/vendir/config/lock_config_test.go
```diff
@@ -0,0 +1,150 @@
+// Copyright 2022 VMware, Inc.
+// SPDX-License-Identifier: Apache-2.0
+
+package config_test
+
+import (
+	"github.com/stretchr/testify/require"
+	"github.com/vmware-tanzu/carvel-vendir/pkg/vendir/config"
+	"testing"
```
joaopapereira:
```suggestion
	"testing"

	"github.com/stretchr/testify/require"
	"github.com/vmware-tanzu/carvel-vendir/pkg/vendir/config"
```

joaopapereira (changes requested):
Thanks for the PR, looks like we are almost there. Please see my comments

pkg/vendir/cmd/sync.go
petewall: Good question. I thought about where to put the check, because essentially, "if it's the same, don't write it" means we need to know three things:
1. Does the lock file even exist? Don't blow up checking an "old" lock file that doesn't exist
2. Did the lock file change? this means we need to read it sometime
3. Is the desired state of the lock file done updating?

Twice in the sync command we read the lock file. First in the `if o.Locked` block and once in the `if len(dirs) > 0` block near the end. I wanted to piggy-back on those if possible.

Maybe I'm being too clever with it and adding confusion. Let me try something simpler and see what you think.

petewall: New change incoming...

test/e2e/example.go
```diff
@@ -113,11 +113,25 @@ func (et example) check(t *testing.T, vendir Vendir) error {
 		}
 	}
 
+	lockFileStat, err := os.Stat(filepath.Join(path, "vendir.lock.yml"))
+	if err != nil {
+		return fmt.Errorf("Expected no err for getting lock file stats")
+	}
+
```
joaopapereira:
```suggestion
	require.NoError(t, err)
```

test/e2e/example.go
```diff
@@ -113,11 +113,25 @@ func (et example) check(t *testing.T, vendir Vendir) error {
...
 	_, err = vendir.RunWithOpts([]string{"sync", "--locked"}, RunOpts{Dir: path, Env: et.Env})
 	if err != nil {
 		return fmt.Errorf("Expected no err for sync locked")
 	}
 
+	newLockFileStat, err := os.Stat(filepath.Join(path, "vendir.lock.yml"))
+	if err != nil {
+		return fmt.Errorf("Expected no err for getting lock file stats")
+	}
```
joaopapereira:
```suggestion
	require.NoError(t, err)
```

test/e2e/example.go
```diff
@@ -113,11 +113,25 @@ func (et example) check(t *testing.T, vendir Vendir) error {
...
+
+	if lockFileStat.ModTime() != newLockFileStat.ModTime() {
+		return fmt.Errorf("Expected lock file to not be updated")
+	}
```
joaopapereira:
```suggestion
	require.Equal(t, lockFileStat.ModTime(), newLockFileStat.ModTime())
```

joaopapereira (approved):
LGTM

cppforlife (approved):

PR #191, opened 2022-10-11
Check on existence of unpackArchive path 

### LINKED ISSUE #184: clearer error message if path for githubrelease archive does not exist
**What steps did you take:**
I tried to unpack a github release with the wrong archive name. Here is a vendir.yaml file to replicate:
```
kind: Config
directories:
- path: vendor
  contents:
  # unpacks archive included in the release
  - path: github.com/cloudfoundry-incubator/eirini-release
    githubRelease:
      slug: cloudfoundry-incubator/eirini-release
      tag: v1.2.0
      checksums:
        eirini-cf.tgz: 819b37126f81ad479acc8dcd7e61e8b0e55153d8fa27aa9a04692c38d0c310fe
        eirini-uaa.tgz: efe8a498c67368fac1c46fa52c484261cbf4e78b3291cff69e660cd863342674
        eirini.tgz: b535d9434300e79d11d42acc417148d09054aa32808dab5f12264d3af59ad548
      unpackArchive:
        path: thisfiledoesnotexist.tgz
```

**What happened:**
You get the error:
> vendir: Error: Syncing directory 'vendor':
  Syncing directory 'github.com/cloudfoundry-incubator/eirini-release' with github release contents:
    Expected known archive type (zip, tgz, tar)

**What did you expect:**
An error message which tells me that the specified file does not exist. Can be implemented by a check before trying to unpack it.

**Environment:**

- vendir version: 0.31.0
- OS: Debian Bookworm

---
Vote on this request

kumaritanushree:
Added check on existence of `unpackArchive` path and throw detailed error if path does not found.
Fixes  #184

### DIFF
--- a/pkg/vendir/fetch/githubrelease/sync.go
+++ b/pkg/vendir/fetch/githubrelease/sync.go
@@ -175,6 +175,14 @@ func (d Sync) Sync(dstPath string, tempArea ctlfetch.TempArea) (ctlconf.LockDire
 
 		defer os.RemoveAll(newIncomingTmpPath)
 
+		_, err = os.Stat(filepath.Join(incomingTmpPath, d.opts.UnpackArchive.Path))
+		if err != nil {
+			if os.IsNotExist(err) {
+				return lockConf, fmt.Errorf("Unpacking archive '%s' is not part of the github release", d.opts.UnpackArchive.Path)
+			}
+			return lockConf, err
+		}
+
 		final, err := ctlfetch.NewArchive(filepath.Join(incomingTmpPath, d.opts.UnpackArchive.Path), false, "").Unpack(newIncomingTmpPath)
 		if err != nil {
 			return lockConf, fmt.Errorf("Unpacking archive '%s': %s", d.opts.UnpackArchive.Path, err)
--- a/test/e2e/github_release_test.go
+++ b/test/e2e/github_release_test.go
@@ -1,0 +1,77 @@
+// Copyright 2022 VMware, Inc.
+// SPDX-License-Identifier: Apache-2.0
+
+package e2e
+
+import (
+	"os"
+	"strings"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func TestGithubRelease(t *testing.T) {
+	env := BuildEnv(t)
+	logger := Logger{}
+	vendir := Vendir{t, env.BinaryPath, logger}
+
+	dstPath, err := os.MkdirTemp("", "vendir-e2e-github-release-dst")
+	require.NoError(t, err)
+
+	defer os.RemoveAll(dstPath)
+
+	yaml := `
+apiVersion: vendir.k14s.io/v1alpha1
+kind: Config
+directories:
+- path: vendor
+  contents:
+  # unpacks archive included in the release
+  - path: github.com/cloudfoundry-incubator/eirini-release
+    githubRelease:
+      slug: cloudfoundry-incubator/eirini-release
+      tag: v1.2.0
+      checksums:
+        eirini-cf.tgz: 819b37126f81ad479acc8dcd7e61e8b0e55153d8fa27aa9a04692c38d0c310fe
+        eirini-uaa.tgz: efe8a498c67368fac1c46fa52c484261cbf4e78b3291cff69e660cd863342674
+        eirini.tgz: b535d9434300e79d11d42acc417148d09054aa32808dab5f12264d3af59ad548
+      unpackArchive:
+        path: thisfiledoesnotexist.tgz
+`
+	yaml1 := `
+apiVersion: vendir.k14s.io/v1alpha1
+kind: Config
+directories:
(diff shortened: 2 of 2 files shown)


pkg/vendir/fetch/githubrelease/sync.go
```diff
@@ -175,6 +175,11 @@ func (d Sync) Sync(dstPath string, tempArea ctlfetch.TempArea) (ctlconf.LockDire
 
 		defer os.RemoveAll(newIncomingTmpPath)
 
+		_, err = os.Stat(filepath.Join(incomingTmpPath, d.opts.UnpackArchive.Path))
+		if err != nil {
+			return lockConf, err
```
joaopapereira, line 183: I think we should be a little bit more explicit on the error message. If it is PathNotFound we should let people know that the file they want to unpack is not part of the github release

joaopapereira (changes requested):
I added a comment about the error message.
Also, do you think adding some tests around this behavior is possible?

pkg/vendir/fetch/githubrelease/sync.go
kumaritanushree: Updated error with more information and added e2e test as well.

test/e2e/github_release_test.go
```diff
@@ -0,0 +1,74 @@
...
+        path: eirini.tgz
+`
+
+	expectedErr := `vendir: Error: Syncing directory 'vendor':
+  Syncing directory 'github.com/cloudfoundry-incubator/eirini-release' with github release contents:
+    Unpacking archive 'thisfiledoesnotexist.tgz' is not part of the github release
+`
+
+	logger.Section("sync with mismatch github release path", func() {
+		_, err := vendir.RunWithOpts([]string{"sync", "-f", "-"}, RunOpts{StdinReader: strings.NewReader(yaml), AllowError: true})
```
joaopapereira: I think we should use a temporary directory to run these tests on. If not we will be polluting the folder where the test file is present

joaopapereira (changes requested):
The implementation looks good, please just ensure you are using a temp directory to store the results of the test

test/e2e/github_release_test.go
```diff
@@ -0,0 +1,77 @@
...
+
+func TestGithubRelease(t *testing.T) {
+	env := BuildEnv(t)
+	logger := Logger{}
+	vendir := Vendir{t, env.BinaryPath, logger}
+
+	dstPath, err := os.MkdirTemp("", "vendir-e2e-github-release-dst")
+	if err != nil {
+		t.Fatal(err)
+	}
```
joaopapereira:
```suggestion
	require.NoError(t, err)
```

This change is just to keep the pattern we are using consistent

test/e2e/github_release_test.go
```diff
@@ -0,0 +1,77 @@
...
+// SPDX-License-Identifier: Apache-2.0
+
+package e2e
+
+import (
+	"os"
+	"strings"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
```
joaopapereira, line 11:
```suggestion
	"github.com/stretchr/testify/assert"
	"github.com/stretchr/testify/require"
```

test/e2e/github_release_test.go
```diff
@@ -0,0 +1,77 @@
...
+`
+
+	expectedErr := `vendir: Error: Syncing directory 'vendor':
+  Syncing directory 'github.com/cloudfoundry-incubator/eirini-release' with github release contents:
+    Unpacking archive 'thisfiledoesnotexist.tgz' is not part of the github release
+`
+
+	logger.Section("sync with mismatch github release path", func() {
+		_, err := vendir.RunWithOpts([]string{"sync", "-f", "-"}, RunOpts{Dir: dstPath, StdinReader: strings.NewReader(yaml), AllowError: true})
+		assert.Error(t, err, "Expected to err while syncing with github release")
```
joaopapereira:
```suggestion
		require.Error(t, err, "Expected to err while syncing with github release")
```

test/e2e/github_release_test.go
```diff
@@ -0,0 +1,77 @@
...
+		assert.ErrorContains(t, err, expectedErr)
+	})
+
+	logger.Section("sync again with github release", func() {
+		_, err := vendir.RunWithOpts([]string{"sync", "-f", "-"}, RunOpts{Dir: dstPath, StdinReader: strings.NewReader(yaml1)})
+		assert.NoError(t, err)
```
joaopapereira:
```suggestion
		require.NoError(t, err)
```

joaopapereira (changes requested):
A couple of comments.

When you use the assert package in testify even if the assertion fails the test continues and in cases where we want to check for error we should always use require because if the assertion fails the test will stop. Usually, there is no good reason to continue a test when we are checking for the existence or absence of an error and that assertion fails.

joaopapereira (approved):
LGTM

cppforlife (approved):

PR #199, opened 2022-12-02
Make directory permissions configurable

hoegaarden:
- Inner directories inherit permissions from the outer directory
- the permissions on the directories is changed late, so that pulling in content is not blocked

/xref: https://github.com/vmware-tanzu/carvel-vendir/issues/186

### DIFF
--- a/test/e2e/directory_permissions_test.go
+++ b/test/e2e/directory_permissions_test.go
@@ -35,7 +35,7 @@ func TestDirectoryPermissions(t *testing.T) {
 				c.Directories[0].Permission = p(0744)
 			},
 			expectedPerms: filePerms{
-				"dir-0": 0744, filepath.Join("dir-0", "subdir-0-0"): 0700, filepath.Join("dir-0", "subdir-0-1"): 0700,
+				"dir-0": 0744, filepath.Join("dir-0", "subdir-0-0"): 0744, filepath.Join("dir-0", "subdir-0-1"): 0744,
 				"dir-1": 0700, filepath.Join("dir-1", "subdir-1-0"): 0700, filepath.Join("dir-1", "subdir-1-1"): 0700,
 			},
 		},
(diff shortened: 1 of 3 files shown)


cppforlife:
cc @joaopapereira you previously had some thoughts on permissions. any objections to this PR?

test/e2e/directory_permissions_test.go
```diff
@@ -0,0 +1,171 @@
...
+package e2e
+
+import (
+	"os"
+	"path/filepath"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/vmware-tanzu/carvel-vendir/pkg/vendir/config"
+	"gopkg.in/yaml.v2"
```
joaopapereira: We are using this library inside the code, maybe it makes sense to keep it consistent. I just realized that the gopkg one is used in a different tests, I will create a PR to fix that as well
```suggestion
	"sigs.k8s.io/yaml"
```

test/e2e/directory_permissions_test.go
```diff
@@ -0,0 +1,171 @@
...
+	for path, expectedPerms := range expected {
+		actualPerms, ok := actual[path]
+		assert.True(t, ok, "no actual permissions for path %s found", path)
+		assert.Equal(t, expectedPerms, actualPerms, "expected permissions for '%s' to be '%s', but got '%s'", path, expectedPerms, actualPerms)
+	}
+}
+
+func writeConfigFile(t *testing.T, tmpDir string, config config.Config) {
+	vendirYML, err := os.Create(filepath.Join(tmpDir, "vendir.yml"))
+	assert.NoError(t, err, "creating vendir.yml")
```
joaopapereira: I think we should use require in this place as well as in other parts. The main difference between `require` and `assert` is that assert allows the these to continue, but in situations like this we should not continue with the execution of the test. (I did not highlight all the places where this change can be done)
```suggestion
	require.NoError(t, err, "creating vendir.yml")
```

joaopapereira (changes requested):
Hey, thanks for putting together this PR.

The PR implementation looks good, I added a couple of comments mainly on the test side of the world.
Also when you finish this do you mind creating a PR in https://github.com/vmware-tanzu/carvel to update the documentation with the new permission field? The website is organized by version and your changes should be inside [this directory](https://github.com/vmware-tanzu/carvel/tree/develop/site/content/vendir/docs/develop)

joaopapereira (approved):
LGTM

cppforlife (commented):
lets rename field name to be plural: permission => permissions

PR #210, opened 2023-01-17
Enable imgpkg keychains when environment variable is provided

### LINKED ISSUE #209: vendir doesn't load correct keychains when provided with the environment variables
**What steps did you take:**
When sync an image while trying to use the Credentials Helpers from the Azure/AWS/GCP are failing with 401 error

**What happened:**
Fails to pull images using the Credentials helpers

**What did you expect:**
That vendir would be able to use the cred helpers from imgpkg to download the images

The environment variable `IMGPKG_ACTIVE_KEYCHAINS` is not being respected because it is not taken into account when creating the registry options in https://github.com/vmware-tanzu/carvel-vendir/blob/develop/pkg/vendir/fetch/image/imgpkg.go#L157

**Environment:**

- vendir version (execute `vendir --version`): latest version
- OS (e.g. from `/etc/os-release`): all

---
Vote on this request

We are also happy to receive and review Pull Requests if you want to help work on this issue.

joaopapereira:
Fixes #209


pkg/vendir/fetch/image/imgpkg.go
```diff
@@ -154,11 +156,30 @@ func (t *Imgpkg) RegistryOpts() (registry.Opts, error) {
...
+				return registry.Opts{}, fmt.Errorf("Expected 'IMGPKG_ACTIVE_KEYCHAINS' environment variable to have a list of keychains but got '%s'", envVar)
+			}
+
+			if strings.Contains(keychains[1], ",") {
+				for _, keychainName := range strings.Split(keychains[1], ",") {
+					activeKeychains = append(activeKeychains, auth.IAASKeychain(strings.TrimSpace(keychainName)))
+				}
+			} else {
+				activeKeychains = append(activeKeychains, auth.IAASKeychain(strings.TrimSpace(keychains[1])))
+			}
```
praveenrewar: *nit*: I think the result will be the same even if we remove the `if else` and just keep the for loop 🤔

pkg/vendir/fetch/image/imgpkg.go
```diff
@@ -154,11 +156,30 @@ func (t *Imgpkg) RegistryOpts() (registry.Opts, error) {
 		return registry.Opts{}, err
 	}
 
+	var activeKeychains []auth.IAASKeychain
```
cppforlife: should this logic be included into imgpkg itself (e.g. within registry.Opts constructor for example). it seems that we are duplicating functionality. wdyt?

joaopapereira: maybe we could do it with all the logic, I will look into it

cppforlife (approved):

PR #235, opened 2023-03-02
Bump go version to 1.20.1

rohitagg2020:
- Bump go version to 1.20.1
- Bump golangci-lint to 1.51.2 as go 1.20 support was added in 1.51.0
- Fixing golangci-lint error: rand.Seed has been deprecated since Go 1.20. 
            Seeding is added by default from 1.20 onwards. More info [here]( https://github.com/golang/go/blob/3f8f929d60a90c4e4e2b07c8d1972166c1a783b1/src/math/rand/rand.go#L383)


joaopapereira (approved):
LGTM

PR #239, opened 2023-03-16
fix: Add validation to not fail on same secret

### LINKED ISSUE #238: Sync fails when providing the same secret more than once
**What steps did you take:**
[A clear and concise description steps that can be used to reproduce the problem.]

1. Create a private git repository with multiple folders to sync

```bash
repo
|- app
|--- app.yaml
|- cm
|--- cm.yaml
|- secret
|--- secret.yaml
```

3. Create the `vendir.yml` file with multiple `content` entries, each to different folder of the same repo, each entry has the `secretRef` with the same _Secret_ `name`

```yaml
kind: Config
directories:
- path: "repo"
  contents:
  - path: cm
    git:
      url: git@github.com:odinnordico/kapp-controller-private-test.git
      secretRef:
          name: ssh-key-secret
      ref: origin/main
  - path: secret
    git:
      url: git@github.com:odinnordico/kapp-controller-private-test.git
      secretRef:
          name: ssh-key-secret
      ref: origin/main
---
apiVersion: v1
data:
  ssh-privatekey: LS...o=
kind: Secret
metadata:
  creationTimestamp: null
  name: ssh-key-secret
---
apiVersion: v1
data:
  ssh-privatekey: LS...=
kind: Secret
metadata:
  creationTimestamp: null
  name: ssh-key-secret
---
apiVersion: v1
data:
  ssh-privatekey: LS...o=
kind: Secret
metadata:
  creationTimestamp: null
  name: ssh-key-secret
```

5. Run the vendir command
```bash
vendir sync 
```
6. Check the output:

```bash
Fetching: repo + app (git from git@github.com:odinnordico/kapp-controller-private-test.git@origin/main)

vendir: Error: Syncing directory 'repo':
  Syncing directory 'app' with git contents:
    Fetching git repository:
      Expected to find one secret 'ssh-key-secret', but found multiple
```

**What happened:**

When using the same secret to sync folders in the same private repo and providing the secret more than once, `vendir` fails expecting only one secret and not matter if it is the same

**What did you expect:**
[A description of what was expected]

The directory syncs correctly with the private git repo

[Additional information that will assist in solving the issue.]

Kapp Controller related issue: [1002](https://github.com/carvel-dev/kapp-controller/issues/1002)

**Environment:**

- vendir version (execute `vendir --version`): vendir version 0.33.1

PRETTY_NAME="Debian GNU/Linux bookworm/sid"
NAME="Debian GNU/Linux"
VERSION_CODENAME=bookworm
ID=debian

---
Vote on this request

odinnordico:
Add validation when getting the secret to not fail if more than two secrets with the same `name` and same `data` are provided in the config yaml.

Fixes #238 

**Test output:**

```bash
VENDIR_GITHUB_API_TOKEN=ghp_*** ./hack/build.sh && ./hack/test-all.sh

vendir version 0.0.0+develop
+ ./vendir version
vendir version 0.0.0+develop

Succeeded
+ go test --exec=echo ./...
+ echo Success
Success
...
+ echo UNIT SUCCESS
UNIT SUCCESS
...
+ echo E2E SUCCESS
E2E SUCCESS
+ echo ALL SUCCESS
ALL SUCCESS
```

**Local Test:**

_vendir.yml_

```yaml
kind: Config
directories:
- path: "repo"
  contents:
  - path: "1"
    git:
      url: git@github.com:odinnordico/kapp-controller-private-test.git
      secretRef:
          name: ssh-key-secret
      ref: origin/main
    includePaths:
    - cm/**/*
  - path: "2"
    git:
      url: git@github.com:odinnordico/kapp-controller-private-test.git
      secretRef:
          name: ssh-key-secret
      ref: origin/main
    includePaths:
    - secret/**/*
---
apiVersion: v1
data:
  ssh-privatekey: LS---o=
kind: Secret
metadata:
  name: ssh-key-secret
---
apiVersion: v1
data:
  ssh-privatekey: LS...o=
kind: Secret
metadata:
  name: ssh-key-secret
---
apiVersion: v1
data:
  ssh-privatekey: LS...o=
kind: Secret
metadata:
  name: ssh-key-secret

```

_Output_

```bash
Fetching: repo + 1 (git from git@github.com:odinnordico/kapp-controller-private-test.git@origin/main)

  --> git init
  Initialized empty Git repository in /home/diego/workspace/test/vendir-secrets/.vendir-tmp/incoming/git/.git/
  --> git config credential.helper store --file /home/diego/workspace/test/vendir-secrets/.vendir-tmp/incoming/git-auth/.git-credentials
  --> git remote add origin git@github.com:odinnordico/kapp-controller-private-test.git
  --> git config remote.origin.tagOpt --tags
  --> git fetch origin
  From github.com:odinnordico/kapp-controller-private-test
   * [new branch]      main       -> origin/main
  --> git -c advice.detachedHead=false checkout origin/main
  HEAD is now at f7c4e58 Fix secret
  --> git submodule update --init --recursive
  --> git rev-parse HEAD
  f7c4e58e3e3a1d7b1180e6f8288a76746b1bea8c
  --> git describe --tags f7c4e58e3e3a1d7b1180e6f8288a76746b1bea8c
  fatal: No names found, cannot describe anything.
  --> git log -n 1 --pretty=%B f7c4e58e3e3a1d7b1180e6f8288a76746b1bea8c
  Fix secret

Fetching: repo + 2 (git from git@github.com:odinnordico/kapp-controller-private-test.git@origin/main)

  --> git init
  Initialized empty Git repository in /home/diego/workspace/test/vendir-secrets/.vendir-tmp/incoming/git/.git/
  --> git config credential.helper store --file /home/diego/workspace/test/vendir-secrets/.vendir-tmp/incoming/git-auth/.git-credentials
  --> git remote add origin git@github.com:odinnordico/kapp-controller-private-test.git
  --> git config remote.origin.tagOpt --tags
  --> git fetch origin
  From github.com:odinnordico/kapp-controller-private-test
   * [new branch]      main       -> origin/main
  --> git -c advice.detachedHead=false checkout origin/main
  HEAD is now at f7c4e58 Fix secret
  --> git submodule update --init --recursive
  --> git rev-parse HEAD
  f7c4e58e3e3a1d7b1180e6f8288a76746b1bea8c
  --> git describe --tags f7c4e58e3e3a1d7b1180e6f8288a76746b1bea8c
  fatal: No names found, cannot describe anything.
  --> git log -n 1 --pretty=%B f7c4e58e3e3a1d7b1180e6f8288a76746b1bea8c
  Fix secret

Lock config

directories:
- contents:
  - git:
      commitTitle: Fix secret
      sha: f7c4e58e3e3a1d7b1180e6f8288a76746b1bea8c
    path: "1"
  - git:
      commitTitle: Fix secret
      sha: f7c4e58e3e3a1d7b1180e6f8288a76746b1bea8c
    path: "2"
  path: repo
kind: LockConfig

Succeeded
```

_Tree_

```bash
tree .                                                                                                                
.
├── repo
│   ├── 1
│   │   └── cm
│   │       └── cm.yaml
│   └── 2
│       └── secret
│           └── secret.yaml
├── vendir.lock.yml
└── vendir.yml

6 directories, 4 files
```

**Additional Notes**

- The final structure is the result of the `vendir.yml` config, not intendet to be different for this example.
- e2e changed not in git secret but in other usage of `GetSecret` function to check impact


pkg/vendir/directory/named_ref_fetcher.go
```diff
@@ -25,6 +26,9 @@ func (f NamedRefFetcher) GetSecret(name string) (ctlconf.Secret, error) {
 	var found []ctlconf.Secret
 	for _, secret := range f.secrets {
 		if secret.Metadata.Name == name {
+			if len(found) == 1 && reflect.DeepEqual(found[0].Data, secret.Data) {
```
cppforlife: code wouldnt work for 3 secrets that are the same.

odinnordico: The test was done with 3 secrets, if the second secret has the same `data` as the first one, it won't be added to the `found` array, if the third one has the same `data`, won't be added as well, of one of the secrets with same name has different `data` it will fail with multiple secrets error due to we would not assume which one is good to go

cppforlife: you are right. [secret A, secret B, secret B] wont work though, since you are hard coding index lookups.

odinnordico: Yes, that is the intent, if there are more than one secret with same `name` but different `data`, we should fail as we do not know which `secret.data` to use, in that case we show the multiple secrets found error

neil-hickey: That seems reasonable to me, I think in cases where there is obscurity - we should defer to the user.

`[secret A, secret B, secret B]` - When you mentioned this @cppforlife , did you have an outcome you were expecting instead of an explicit error from vendir?

neil-hickey: I get why the test here was put into an e2e situation (git). But the behaviour we are testing can be done via a unit test. (NOTE: This is a simple test, I didn't go through each of the potential problems -- just giving this as an example to work from) :

pkg/vendir/directory/named_ref_fetcher_test.go
```
// Copyright 2023 VMware, Inc.
// SPDX-License-Identifier: Apache-2.0

package directory

import (
	"testing"

	"github.com/vmware-tanzu/carvel-vendir/pkg/vendir/config"
)

func TestMultipleSecret_SameName(t *testing.T) {
	secrets := []config.Secret{
		{
			Metadata: config.GenericMetadata{
				Name: "secret-a",
			},
			Data: map[string][]byte{
				"foo": []byte("bar"),
			},
		},
		{
			Metadata: config.GenericMetadata{
				Name: "secret-b",
			},
			Data: map[string][]byte{
				"foo": []byte("bar"),
			},
		},
		{
			Metadata: config.GenericMetadata{
				Name: "secret-b",
			},
			Data: map[string][]byte{
				"foo": []byte("bar"),
			},
		},
	}

	nf := NewNamedRefFetcher(secrets, []config.ConfigMap{})
	_, err := nf.GetSecret("secret-b")

	require.NoError(t, err)
}

```

cppforlife: > [secret A, secret B, secret B] - When you mentioned this @cppforlife , did you have an outcome you were expecting instead of an explicit error from vendir?

im saying the way it's currently implemented, it only deals with a single secret being duplicated, and it has to come in a very specific order (dup-ed secrets must come first). neither of these assumptions should be there. we should be able to support N numbers of dup-ed secrets in any order. to support that, just create a separate map keyed by secret name and do the lookup (dont use arrays for keyed lookups)

odinnordico: @cppforlife i am not sure if i am understanding right your concern,
the function only cares of secret by the name, that is why i am comparing only that, and if the `len(found)` is `0` i add the secret, if it is already a secret in `found` with same `data`, we do not need to add it but if there is one secret in `found` (meaning with the name that we care about) but data is different then we add it and later one the `len` validation will throw the error, if there is more than one secrets in `found` then we do no care on validating `data` it will fails due to more than one secret found

neil-hickey: Perhaps having a test case that fails would be useful here to clear things up? I think in this case being really concrete with a failure will help guide the implementation.

odinnordico: i just added the test file with two cases failing (`wantErr: true`)

neil-hickey: @cppforlife ^ tests were added. If they don't cover what you were expecting, can you add a suggestion for a test that does fail?

cppforlife: sorry folks, i screwed up -- i misread and assumed this was a change in a different piece of code [0] that was checking _all_ secrets at the load time, vs doing it on the fly when secret is requested and filtered by name (so that's why `found[0]` is ok). so with current impl "invalid" secret configs wouldn't be errored on until someone uses it. ideally we would check for this condition in [0], but i know yall already put some effort in this area.

[0] https://github.com/carvel-dev/vendir/blob/develop/pkg/vendir/config/config.go#L30

odinnordico: It seems to me that is a different issue, correct me if I am wrong, one if allowing the same secret (equal `name` and `data`) multiple times and the other one is checking if secrets are yaml valid, do you want those two checks to happen at the same time (in `NewConfigFromFiles` func)? @cppforlife

cppforlife: what i was thinking for NewConfigFromFiles is to never return duplicate secrets. either dup secrets are exactly the same and so we can safely dedup (and only one copy would be returned), or dup secrets are different in content and so error can be returned from NewConfigFromFiles since configuration is not valid. (named look up can stay as simple as it is today just assuming that all names are unique).

odinnordico: Changed the validation to be on `NewConfigFromFiles`

pkg/vendir/config/config.go
```diff
@@ -48,7 +49,17 @@ func NewConfigFromFiles(paths []string) (Config, []Secret, []ConfigMap, error) {
 			if err != nil {
 				return fmt.Errorf("Unmarshaling secret: %s", err)
 			}
-			secrets = append(secrets, secret)
+
+			if s, ok := secretsNames[secret.Metadata.Name]; ok {
+				if len(s) == 1 {
```
cppforlife: nit: no need to check on length since `[]Secret{secret}` is used below

test/e2e/git_test.go
```diff
@@ -63,7 +70,7 @@ directories:
       verification:
         publicKeysSecretRef:
           name: git-pubs
-`, encodedPubKeys, repoPath, ref))
+`, encodedPubKeys, encodedPubKeys, repoPath, ref))
```
cppforlife: no need to modify this test since it's not directly related to this feature (and unit tests should give us enough confidence in this case).

pkg/vendir/config/config_test.go
```diff
@@ -27,3 +27,232 @@ kind: Config`)
...
+  ssh-privatekey: BAZ=
+kind: Secret
+metadata:
+  name: another-secret
+`)
+
+		require.NoError(t, os.WriteFile(tempConfigPath, configWithWhitespace, 0666))
+
+		_, _, _, err := config.NewConfigFromFiles([]string{tempConfigPath})
+		require.Error(t, err)
```
cppforlife, line 258: lets add an err message assertion here

odinnordico: Added assert on the message

test/e2e/git_test.go
odinnordico: Removed from e2e

pkg/vendir/config/config.go
odinnordico: Removed, also not used array as it will fail if multiple secret with same `name` and different `data`

cppforlife (approved):

PR #247, opened 2023-04-07
Removing example which contains AGPL license

kumaritanushree:
Removing example which contains AGPL license.


joaopapereira (changes requested):
I think we need to do a little bit more in this PR it is not enough to remove we should also add a new example to ensure we have some coverage for hg

praveenrewar:
The idea for long term was that we would move all the examples from our repositories to one common repository as the examples lead to a lot of false positive alerts. For now, if we do not want to remove the example, maybe we can use `newRootPath` to not include files from the root directory?

joaopapereira (approved):
LGTM

PR #251, opened 2023-04-27
Avoid fetching unnecessary git tags and branches

scothis:
When a git ref is specified, it's not necessary to pull refs for every tag and branch in the repo. For large repos with lots of branch and tags this change can result in a multiple orders of magnitude speed up.

### DIFF
--- a/pkg/vendir/fetch/git/git.go
+++ b/pkg/vendir/fetch/git/git.go
@@ -147,11 +147,19 @@ func (t *Git) fetch(dstPath string, tempArea ctlfetch.TempArea) error {
 		{"init"},
 		{"config", "credential.helper", "store --file " + gitCredsPath},
 		{"remote", "add", "origin", gitURL},
-		{"config", "remote.origin.tagOpt", "--tags"},
+	}
+
+	if t.opts.RefSelection != nil {
+		// fetch tags for selection
+		argss = append(argss, []string{"config", "remote.origin.tagOpt", "--tags"})
 	}
 
 	{
 		fetchArgs := []string{"fetch", "origin"}
+		if strings.HasPrefix(t.opts.Ref, "origin/") {
+			// only fetch the exact ref we're seeking
+			fetchArgs = append(fetchArgs, t.opts.Ref[7:])
+		}
 		if t.opts.Depth > 0 {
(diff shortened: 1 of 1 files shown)


scothis:
I believe this to be an optimization that is compatible with existing usage patterns, but I'm not an expert in all the ways git repos are consumed with vendir.

Using vendir 0.33.1 fetching a branch took 93 seconds. With this change, syncing the same content took 17 seconds.

neil-hickey:
LGTM! Makes a lot of sense. I can't see a reason this would be a breaking change.... By specifying a branch / tag, you are effectively only ever going to checkout that code anyways. I guess it's possible someone is relying on the fact we are pulling down all the branches locally. But it's not really the intended outcome from specifying a tag. At least I as a user wouldn't expect to rely on this behaviour...

scothis:
@neil-hickey what are the next steps?

kumaritanushree:
@scothis did you tested your changes with ref: <git_tag> . In case of tag it is still fetching all branches and tags. Is it expected? I think you would want to avoid fetching all tags and branch in case of tag as well.

scothis:
@kumaritanushree ~good catch, it works with branches but not tags. I'll keep digging to find a way the makes both happy.~

*edit*: I had misconfigured my vendir.yml for a tag, I tried again and the same vendir.yml works with git branches, tags, and commits with both vendir 0.32.2 and this PR. All of the tags and branches are still fetched for tags and commits. It doesn't take advantage of the optimized flow that exists for branches, but it still works.

kumaritanushree:
@scothis I verified your changes as well and my doubt was do we want same functionality for tag and commit or not?

scothis:
@kumaritanushree sorry for the confusion. It's possible to efficiently clone tags and branches using `git clone -b <branchname | tagname> --depth 1 <repository>`, but this will fail when the ref is a commit digest. We could try to sniff a commit based on the length of the ref, but it won't be perfect.

etirta:
Just want to say I really want this changes so bad. Thx @scothis for improving this. When can this be merged? Cheers.

joaopapereira (approved):
This change looks good, it does have the caveat that it only works for origin-based references.
For now, I think it should be enough because it might be complicated to try to do a more generic solution and the signal we are getting is solved by this work

PR #252, opened 2023-05-03
Bump golang 1.20.3 and golint changes

sethiyash:
- Bump golang to 1.20.3 to fix CVE's in go 1.19.6

### DIFF
--- a/pkg/vendir/directory/staging_dir.go
+++ b/pkg/vendir/directory/staging_dir.go
@@ -104,7 +104,7 @@ func (d StagingDir) CopyExistingFiles(rootDir string, stagingPath string, ignore
 		}
 
 		// Move the file to the staging directory
-		err = copy(path, stagingPath)
+		err = copyFile(path, stagingPath)
 		if err != nil {
 			return fmt.Errorf("Moving source file '%s' to staging location '%s': %s", path, stagingPath, err)
 		}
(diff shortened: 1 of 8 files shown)


pkg/vendir/fetch/ref_fetcher.go
```diff
@@ -27,6 +27,6 @@ func (f SingleSecretRefFetcher) GetSecret(name string) (ctlconf.Secret, error) {
 	return ctlconf.Secret{}, fmt.Errorf("Not found")
 }
 
-func (f SingleSecretRefFetcher) GetConfigMap(name string) (ctlconf.ConfigMap, error) {
+func (f SingleSecretRefFetcher) GetConfigMap(_ string) (ctlconf.ConfigMap, error) {
```
kumaritanushree, line 30: If we are not using this param here then why we are not removing it instead of using `_`?

sethiyash: Here `GetConfigMap` is part of `RefFetcher` interface and we can't change the method signatures.

pkg/vendir/directory/staging_dir.go
```diff
@@ -104,7 +104,7 @@ func (d StagingDir) CopyExistingFiles(rootDir string, stagingPath string, ignore
 		}
 
 		// Move the file to the staging directory
-		err = copy(path, stagingPath)
+		err = copyFunc(path, stagingPath)
```
praveenrewar: Is there a copy function introduced in Go v1.20? 
nit: copyFunc -> copyFile

sethiyash: Not sure when it got introduced but this is what golint is throwing::
```
pkg/vendir/directory/staging_dir.go:197:1: redefines-builtin-id: redefinition of the built-in function copy (revive)
```

praveenrewar (approved):
LGTM!

PR #264, opened 2023-06-08
ensure tags are fetched if a ref is present

### LINKED ISSUE #263: v0.34.0 - Running `vendir sync` fetches outdated repo when pointed at tag
**What steps did you take:**
In my `vendir.yml` file:

```yaml
directories:
- contents:
  - git:
      depth: 1
      ref: v8.291.0 # landed to branch within the last hour
      # [...]
```

ran the following:

```shell
$ vendir sync
```

**What happened:**
```
   * [new tag]           v0.2.0                  -> v0.2.0
   * [new tag]           v8.206.1                -> v8.206.1
  --> git -c advice.detachedHead=false checkout v8.291.0
  error: pathspec 'v8.291.0' did not match any file(s) known to git

vendir: Error: Syncing directory 'vendor':
  Syncing directory 'example' with git contents:
    Fetching git repository:
      Git [-c advice.detachedHead=false checkout v8.291.0]: exit status 1 (stderr: error: pathspec 'v8.291.0"' did not match any file(s) known to git)
```

`v8.206.1` shows up as the most recent tag, although that tag is more than several months old.

**What did you expect:**
`vendir sync` should pull the latest version of the repo.

If I replace :

```yaml
ref: v8.291.0 
```
with:
```yaml
ref: master
```

Then `vendir sync` will pull the latest commit without issue (the commit from which `v8.291.0` was generated), but we prefer to run against tags rather than `master`.

I was able to work around the issue by downloading the `0.33.2` release of vendir and using that to run `vendir sync`, so the problem seems to be contained in `0.34.0`. Since this is the development version, maybe it shouldn't be what Homebrew installs by default? At Ieast splitting it out into development and release channels might be useful?

**Environment:**

- vendir version (execute `vendir --version`): `vendir version 0.34.0` (from Homebrew)
- OS (e.g. from `/etc/os-release`): `macos 13.4 (22F66)`

---
Vote on this request

Laevos:
Fixes #263

### DIFF
--- a/pkg/vendir/fetch/git/git.go
+++ b/pkg/vendir/fetch/git/git.go
@@ -149,7 +149,7 @@ func (t *Git) fetch(dstPath string, tempArea ctlfetch.TempArea) error {
 		{"remote", "add", "origin", gitURL},
 	}
 
-	if t.opts.RefSelection != nil {
+	if t.opts.RefSelection != nil || t.opts.Ref != "" {
 		// fetch tags for selection
 		argss = append(argss, []string{"config", "remote.origin.tagOpt", "--tags"})
 	}


pkg/vendir/fetch/git/git.go
```diff
@@ -149,7 +149,7 @@ func (t *Git) fetch(dstPath string, tempArea ctlfetch.TempArea) error {
 		{"remote", "add", "origin", gitURL},
 	}
```
neil-hickey, line 151: I'd say let's have it look like this instead 

```
        ref, err := t.resolveRef(dstPath)
	if err != nil {
		return err
	}
	
        // fetch tags for selection
	argss = append(argss, []string{"config", "remote.origin.tagOpt", "--tags"})

        {
		fetchArgs := []string{"fetch", "origin"}
		if strings.HasPrefix(ref, "origin/") {
			// only fetch the exact ref we're seeking
			fetchArgs = append(fetchArgs, ref[7:])
		}
		if t.opts.Depth > 0 {
			fetchArgs = append(fetchArgs, "--depth", strconv.Itoa(t.opts.Depth))
		}
		argss = append(argss, fetchArgs)
	}

        .... REMOVE `t.resolveRef(dstPath)` from lines 174-177
```

Without any testing it out, I think this will provide us 

a) the opts != nil guards we want.
b) does nothing in the case you select `RefSelection` -  I don't believe `RefSelection` even resolves to a non-semver version (i.e it won't have origin/ in it)

neil-hickey: Nevermind, we can't list tags first before we fetch them. Scratch this idea.

Laevos: Yeah, I was wondering if there were a way that we could proactively check a single tag, namely whatever's in `Ref`. Maybe something like `git tag -l $(ref)`?

Laevos: But again I guess we'd need to fetch them all first :/

neil-hickey:
So this particular fix just reverts the feature we implemented. So we will submit up a fix that solves for the feature working, and keep it's intended behaviour. Thanks @Laevos

Laevos:
sgtm, thanks folks!


Jun 8, 2023 14:20:31 Neil Hickey ***@***.***>:

So this particular fix just reverts the feature we implemented. So we will submit up a fix that solves for the feature working, and keep it's intended behaviour. Thanks @Laevos<https://github.com/Laevos>

Issue #260, opened 2023-05-31
Feature request: add option for skipping SSL verification when using Git


alexandreroman:
**Describe the problem/challenge you have**
vendir has no option for setting `GIT_SSL_NO_VERIFY=true`, which could be used with Git repos using a self-signed certificate.
[This environment variable tells Git not to verify SSL certificates](https://git-scm.com/book/en/v2/Git-Internals-Environment-Variables).

**Describe the solution you'd like**
The current implementation already sets a few environment variables when using Git, such as `GIT_LFS_SKIP_SMUDGE=1` to [skip downloading lfs files](https://github.com/carvel-dev/vendir/blob/develop/pkg/vendir/fetch/git/git.go#L121).

I wish there was a way to disable SSL verification when using Git, by adding a flag in the Git configuration.

For example:

```yaml
git:
      url: https://github.com/cloudfoundry/cf-k8s-networking
      ref: origin/master
      # skip SSL verification
      skipSslVerification: true
```

---
Vote on this request

This is an invitation to the community to vote on issues, to help us prioritize our backlog. Use the "smiley face" up to the right of this comment to vote.

👍 "I would like to see this addressed as soon as possible"
👎 "There are other more important things to focus on right now"

We are also happy to receive and review Pull Requests if you want to help working on this issue.

neil-hickey:
Thanks for submitting this issue @alexandreroman , would you be interesting in making a PR with the change?

alexandreroman:
I'm not into Go, but I'm happy to help with testing when you get something ready.

vmunishwar:
Response from @alexandreroman - 
The problem is: I ran into this issue with my customer when using kapp-controller (through TAP namespace-provisioner) and a private Git repo with a self-signed certificate. Ultimately we’d need to add a setting in kapp-controller to skip TLS validation for Git repos - maybe using the flag [dangerousSkipTLSVerify](https://carvel.dev/kapp-controller/docs/v0.46.0/controller-config/#controller-configuration-spec)?

As per above message seems like we need to implement this in kapp-controller.

alexandreroman:
Correct: from a user perspective, kapp-controller will have to configure vendir to skip TLS verification (this is my use case). Reusing the existing flag `dangerousSkipTLSVerify` would be great.

vmunishwar:
[...]

PR #265, opened 2023-06-10
add option for skipping SSL verification

vmunishwar:
To fix the issue - https://github.com/carvel-dev/vendir/issues/260

### DIFF
--- a/pkg/vendir/config/directory.go
+++ b/pkg/vendir/config/directory.go
@@ -67,9 +67,10 @@ type DirectoryContentsGit struct {
 	// +optional
 	SecretRef *DirectoryContentsLocalRef `json:"secretRef,omitempty"`
 	// +optional
-	LFSSkipSmudge      bool `json:"lfsSkipSmudge,omitempty"`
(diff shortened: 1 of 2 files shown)


kumaritanushree:
Changes looks good to me. It will be good if we can add test for the same if possible.

varshapm:
@joaopapereira @kumaritanushree - I have tested the changes and they are working fine on a local gitlab server.

vmunishwar:
@joaopapereira @kumaritanushree - I have tested the changes and they are working fine on a local gitlab server created with a self signed certificate

joaopapereira (approved):
LGTM

vmunishwar:
Steps to reproduce locally: 
1. Use docker image at https://github.com/danieleagle/gitlab-https-docker/tree/master to start local git lab server. Create a repo.
2. Create self signed certificate using openssl and configure with the server
3. Test vendir sync - It should give an error 'SSL certificate problem: self signed certificate'
4. Tested the fix by adding skipSslVerfication: true. There is no error and repo gets cloned successfully.

kumaritanushree:
LGTM

PR #269, opened 2023-06-12
Bumping docker and docker/distribution dependency in v0.33.x line

sethiyash:
- Bumping docker and docker/distribution dependency


joaopapereira (approved):
LGTM

PR #279, opened 2023-08-07
Added lazy sync functionality

### LINKED ISSUE #278: Only run sync for stable versions, if vendir.yaml has changed since last sync

We have a rollout mechanism that heavily relies on vendir to pull in upstream workloads from Helm Charts, OCI Images and Git Repositories. Most of these workloads have fixed versions and don't change unless their corresponding vendir.yaml is modified. Yet, currently vendir constrains us to sync everything on every run, which slows down the process unnecessarily.

We would benefit from a feature that runs sync only, if the vendir.yaml was changed after the last sync.  We would activate this feature for all workloads with stable versions.

---
Vote on this request

fritzduchardt:
Implementation for: https://github.com/carvel-dev/carvel/tree/develop/proposals/vendir/001-lazy-synching-on-stable-config
Closes #278


pkg/vendir/cmd/sync.go
```diff
@@ -50,6 +51,7 @@ func NewSyncCmd(o *SyncOptions) *cobra.Command {
 
 	cmd.Flags().StringSliceVarP(&o.Directories, "directory", "d", nil, "Sync specific directory (format: dir/sub-dir[=local-dir])")
 	cmd.Flags().BoolVarP(&o.Locked, "locked", "l", false, "Consult lock file to pull exact references (e.g. use git sha instead of branch name)")
+	cmd.Flags().BoolVar(&o.Eager, "eager", false, "Ignore lazy setting in vendir yml and eagerly fetch all remote content")
```
kumaritanushree: Did we decide to have default functionality of vendir to fetch only if vendir.yaml has changed else ignore?

pkg/vendir/directory/directory.go
```diff
@@ -232,6 +288,8 @@ func (d *Directory) Sync(syncOpts SyncOpts) (ctlconf.LockDirectory, error) {
 	return lockConfig, nil
 }
 
+//
+
```
kumaritanushree: These lines not required

pkg/vendir/cmd/sync.go
fritzduchardt: The proposal is already open. I think the default will remain unchanged, but if you have this new feature enabled via vendir.yaml config, you can opt out by setting the eager flag.

fritzduchardt: @kumaritanushree the default behavior will not change - vendir will always fetch. However, if the user sets lazy to true in the vendir.yaml for a particular content, lazy fetching will apply. The global flag serves the purpose to force a fetch, even if lazy is set in vendir.yaml

kumaritanushree:
@fritzduchardt 
I have gone through the functionality and looks good but flag `--lazy` somehow creating confusion for me. As we say in flag description it will override "lazy" value in directory content configuration.

But if we have vendir.yml:
```
kind: Config
directories:
- path: sponnet
  contents:
  - path: .
    lazy: false
    http:
      url: https://github.com/spinnaker/sponnet/archive/refs/tags/v1.26.0.tar.gz
```
And we run just `vendir sync` on this it will not override this value.  Also it does not make much sense to use such config but if will just look into flag `--lazy` it sounds the same.

Here `--lazy` will only update  "lazy" value false  not vice versa and description does not tell that.
I am not able to think better description as of now. 
@joaopapereira  any suggestion on this?

Zebradil:
If I read the proposal correctly, `lazy` is `false` by default in the config. It can be explicitly changed to `true`. To disable the "lazy" from CLI, `--lazy=false` has to be used.

So, it wouldn't make much sense to use `lazy: false` in the config because it's the default. Even if it is used, `--lazy=false` won't affect the corresponding directories, they will be synced "non-lazily".

Does that make sense?

joaopapereira:
I think that @Zebradil answer make sense

kumaritanushree:
LGTM

pkg/vendir/directory/directory.go
```diff
@@ -4,7 +4,10 @@
 package directory
 
 import (
+	"crypto/sha256"
+	"encoding/hex"
 	"fmt"
+	"gopkg.in/yaml.v3"
```
joaopapereira: Can we use `sigs.k8s.io/yaml` instead? Just to keep it consistent

go.mod
```diff
@@ -27,6 +27,8 @@ require (
 	sigs.k8s.io/yaml v1.3.0
 )
 
+require gopkg.in/yaml.v3 v3.0.1
```
joaopapereira: When this becomes indirect again, please let us revert this change

pkg/vendir/cmd/sync.go
```diff
@@ -50,6 +51,7 @@ func NewSyncCmd(o *SyncOptions) *cobra.Command {
 
 	cmd.Flags().StringSliceVarP(&o.Directories, "directory", "d", nil, "Sync specific directory (format: dir/sub-dir[=local-dir])")
 	cmd.Flags().BoolVarP(&o.Locked, "locked", "l", false, "Consult lock file to pull exact references (e.g. use git sha instead of branch name)")
+	cmd.Flags().BoolVar(&o.Lazy, "lazy", true, "Override \"lazy\" value in directory content configuration")
```
joaopapereira: Maybe we can change the text here to something like:
"When this flag is set to 'false' it ignores the 'lazy' value in the directory content configuration"
Do you think it is easier to understand what the flag does?

pkg/vendir/directory/directory.go
```diff
@@ -243,3 +278,17 @@ func maybeChmod(path string, potentialPerms ...*os.FileMode) error {
 
 	return nil
 }
+
+func handleLazySync(oldConfigDigest string, newConfigDigest string, fetchLazyGlobalOverride bool, fetchLazy bool) (bool, bool) {
```
joaopapereira: We try to keep functions that are only used in a single struct internal to the structure even if they do not use the structure itself. That way we will know the scope in which we expect these functions to run.
With that in mind do you mind adding this function to the Directory struct?

pkg/vendir/directory/directory_test.go
```diff
@@ -0,0 +1,60 @@
+package directory
+
+import "testing"
+
+func Test_handleLazySync(t *testing.T) {
```
joaopapereira: I am not a fan of testing not exported functions. I understand here that this code would be more complex than testing from the outside. Is there any refactoring that can make this easier?

joaopapereira (changes requested):
I made some comments in the PR. Also, can we add some end-to-end tests to ensure we are doing the correct thing when using the lazy flag?

pkg/vendir/directory/directory.go
fritzduchardt: Sure

pkg/vendir/cmd/sync.go
fritzduchardt: I was thinking along the same lines. It it less vague about the purpose of this option.

pkg/vendir/directory/directory.go
fritzduchardt: Not at all - done

pkg/vendir/directory/directory_test.go
fritzduchardt: No worries, let's remove it and use the e2e test instead

pkg/vendir/directory/directory.go
fritzduchardt: Sure - done

go.mod
fritzduchardt: Ok

test/e2e/example_lazy_test.go
```diff
@@ -0,0 +1,94 @@
+// Copyright 2020 VMware, Inc.
+// SPDX-License-Identifier: Apache-2.0
+
+package e2e
+
+import (
```
joaopapereira, line 6: We do have `require` package as a dependency, and we are using it to do all the assertions in the other tests, you can see an example [here](https://github.com/carvel-dev/vendir/blob/develop/test/e2e/github_release_test.go).
Do you mind refactoring this test to make it similar to the other tests?

joaopapereira (changes requested):
Added a comment about the added test.
After that is solved I think we are good to merge this PR

test/e2e/example_lazy_test.go
fritzduchardt: Not at all - look much more concise now

joaopapereira (approved):
LGTM

pkg/vendir/cmd/sync.go
jessehu: Hi @joaopapereira  This CLI option 'lazy' (default value is true) is really confusing. With vendir 0.37.0, `vendir sync --lazy=true` won't do lazy fetch.  This behavior does not comply with the proposal:
https://github.com/carvel-dev/carvel/tree/develop/proposals/vendir/001-lazy-synching-on-stable-config#other-approaches-considered
图片

This proposal should mean the CLI option --lazy default to **false**, and if --lazy is set to true, vendir will do lazy fetch,
Also this PR defines 'Lazy bool' as a bool instead of `*bool`, so it can not be used to override the CLI option '--lazy' because it always has a default value 'false'.
```
type DirectoryContents struct {
	Path string `json:"path"`
	Lazy bool   `json:"lazy,omitempty"`
```

Zebradil: @jessehu The proposal was:

- to add the `lazy` option to `vendir.yml` to allow skipping sync when the config is not changed,
- to add the `--lazy` CLI argument to temporarily disable the effect of the `lazy` option.

What you reference to is a _considered_ approach that could've satisfied the use case but wasn't accepted.

image

jessehu: I see. Thanks. However this 'other approach considered' makes more sense. The current code behaves a little weird: 

1. vendir sync --lazy=true won't do lazy fetch (--lazy default to true) since it **does not overrites** in lazy setting in vendir.yml
2. vendir sync --lazy=false won't do lazy fetch since it **overrites** in lazy setting in vendir.yml

According to this behaviour, this CLI flag should be named `ignoreLazySetting` or `noLazy` rather than `lazy` (the setting name in verdir.yml)
```
cmd.Flags().BoolVar(&o.Lazy, "lazy", true, "Set to 'false' it ignores the 'lazy' flag in the directory content configuration")
```

Zebradil: The alternative approach is less flexible, as it doesn't allow configuring "laziness" per directory in the `vendir.yml` configuration file.

The selected name of the CLI flag is a little confusing, I agree. It might've been more explicit to use, as you mentioned, `--noLazy` or alike instead of `--lazy=false`.

PR #286, opened 2023-08-11
fix: vendir sync --directory option error

### LINKED ISSUE #277: Failure on specific directory sync with --directory flag
**What steps did you take:**

This is my vendir.yml
```
kind: Config
directories:
- path: charts/grafana
  contents:
  - path: .
    helmChart:
      name: grafana
      version: 5.5.7
      repository:
        url: https://charts.helm.sh/stable
```
I am running selective directory syncing. Of course, in my example this is not necessary, but my vendir.yml could contain multiple directories.

`vendir sync -d charts/grafana`

**What happened:**

On first sync, when there is no lock file written to my local dist, I get the following error:

`vendir: Error: Reading lock config 'vendir.lock.yml': open vendir.lock.yml: no such file or directory`

Syncing somewhat worked, but no lock file was written.

**What did you expect:**

I would expect no error message, plus a lock file containing the version of the directory that I have synced.

**Environment:**

- vendir version (execute `vendir --version`): 0.29.0
- OS (e.g. from `/etc/os-release`): Ubuntu 20.04

---
Vote on this request

fritzduchardt:
Fixes #277


pkg/vendir/config/lock_config.go
```diff
@@ -53,6 +53,13 @@ func NewLockConfigFromBytes(bs []byte) (LockConfig, error) {
 	return config, nil
 }
 
+func LockFileExists(path string) bool {
+	if _, err := os.Stat(path); err != nil {
+		return false
+	}
+	return true
+}
+
```
Zebradil: There is nothing specific to a config in this function. I'd move it to a more generic utils file (if such exists).

Otherwise, it is worth checking for the exact error, because there [could be several of them](https://pkg.go.dev/os#pkg-variables).

```suggestion
func FileExists(path string) (bool, error) {
	_, err := os.Stat(path)
	if err == nil {
		return true, nil
	}
	if errors.Is(err, fs.ErrNotExist) {
		return false, nil
	}
	return false, err
}

```

fritzduchardt: Fair point - there is not util at this point. Made it part of the SyncOptions struct to keep usage simple, but give it a context specific home.

Zebradil: Don't forget to check for the exact error :-)

fritzduchardt: You got it

kumaritanushree:
LGTM

pkg/vendir/config/lock_config.go
```diff
@@ -53,6 +53,16 @@ func NewLockConfigFromBytes(bs []byte) (LockConfig, error) {
 	return config, nil
 }
 
+func LockFileExists(path string) bool {
```
joaopapereira: Is this function used somewhere?

pkg/vendir/config/lock_config.go
```diff
@@ -114,42 +124,51 @@ func (c LockConfig) FindContents(dirPath, conPath string) (LockDirectoryContents
...
+			return true
 		}
 	}
 
-	if !matched {
-		return fmt.Errorf("Expected to match exactly one directory, but did not match any")
+	return false
+}
+
+func (c LockConfig) AppendContents(path string, appendCon LockDirectoryContents) bool {
```
joaopapereira: Can we update this function to be `func (c *LockConfig) AppendContents(path string, appendCon LockDirectoryContents) bool` so that it shows that we are changing the lock configuration

pkg/vendir/config/lock_config.go
```diff
@@ -114,42 +124,51 @@ func (c LockConfig) FindContents(dirPath, conPath string) (LockDirectoryContents
...
+	for i, dir := range c.Directories {
+		if dir.Path == path {
+			c.Directories[i].Contents = append(dir.Contents, appendCon)
+			return true
+		}
 	}
-	return nil
+	return false
+}
+func (c *LockConfig) AppendDirectory(dir LockDirectory) {
```
joaopapereira: I think we can inline this function, unless you think it is easier to read the code with this and we can move the tests to the caller.

pkg/vendir/cmd/sync.go
```diff
@@ -81,6 +88,7 @@ func (o *SyncOptions) Run() error {
 	}
 
 	if len(dirs) > 0 {
+
```
joaopapereira:
```suggestion
```

joaopapereira (changes requested):
Thanks for the PR, I added some comments with some change requests.

pkg/vendir/config/lock_config.go
fritzduchardt: Not really, let's get rid of it

pkg/vendir/config/lock_config.go
fritzduchardt: Sure

pkg/vendir/config/lock_config.go
fritzduchardt: No worries, have inlined and moved the test into the test for Merge

pkg/vendir/cmd/sync.go
fritzduchardt: Done

joaopapereira:
@fritzduchardt looks like a test is failing, also, there are some conflicts. Do you mind checking that?

fritzduchardt:
> @fritzduchardt looks like a test is failing, also, there are some conflicts. Do you mind checking that?

Finally got that fixed. It was harder than expected, since the staging to final location copy logic was being affected.

Previously, the logic for dir-options rewrote the structure of vendir.yml. The staging copy logic relied on this new structure. I took out the structure rewrite, because thanks to my fix, the lock file is now written, even if sync is called with dir-options and we don't want its structure to deviate from the structure of vendir.yml.

fritzduchardt:
@joaopapereira please have another look

joaopapereira:
We are in a good spot, but I will let @kumaritanushree review it.
The only ask for you @fritzduchardt is to get the commits signed so we can get rid of the failing DCO check

fritzduchardt:
> We are in a good spot, but I will let @kumaritanushree review it. The only ask for you @fritzduchardt is to get the commits signed so we can get rid of the failing DCO check

@joaopapereira I squashed them and signed them. Please have another look.

kumaritanushree:
Looks good to me

joaopapereira (approved):
LGTM

PR #307, opened 2023-10-17
Bumping x/net to 0.17.0 in v0.33.x line

sethiyash:


go.mod
```diff
@@ -95,12 +95,12 @@ require (
 	github.com/spf13/pflag v1.0.5 // indirect
 	github.com/vbatts/tar-split v0.11.2 // indirect
 	github.com/vito/go-interact v1.0.1 // indirect
-	golang.org/x/mod v0.7.0 // indirect
-	golang.org/x/net v0.7.0 // indirect
+	golang.org/x/mod v0.8.0 // indirect
+	golang.org/x/net v0.10.0 // indirect
```
praveenrewar: Should be 0.17.0, right?

sethiyash: Thanks for bringing it up. Looks like it got missed somehow.

PR #318, opened 2023-11-09
Bump go version to 1.20.11 in v0.34.x line and imgpkg to v0.37.8

sethiyash:


rohitagg2020 (commented):
Should we be bumping imgpkg version also once it is released?

rohitagg2020 (commented):
LGTM

PR #322, opened 2023-11-10
fix: lazy sync clears vendor directory on second sync

### LINKED ISSUE #319: issue with lazy sync functionality
Background:

The vendir.yml file contains the configuration lazy: true. Upon running vendir sync on a clean project, the project was updated with the vendor directory and all dependencies, along with the vendir.lock.yaml. However, running vendir sync a second time resulted in the complete clearance of the vendor/ directory.

sample vendir.yml :

```
kind: Config
directories:
- path: vendor
  contents:
  - path: dir
    lazy: true
    directory:
      path: ./some-content
```

What I expected:
I expected the vendir sync to skip populating the vendor directory in the second run without clearing the directory.

Environment:

Vendir version: 0.36.0
Operating System: Photon 3/5 or Ubuntu

---
Vote on this request

fritzduchardt:
/closes #319

### DIFF
--- a/pkg/vendir/directory/directory.go
+++ b/pkg/vendir/directory/directory.go
@@ -89,6 +89,11 @@ func (d *Directory) Sync(syncOpts SyncOpts) (ctlconf.LockDirectory, error) {
 		if skipFetching {
 			d.ui.PrintLinef("Skipping fetch: %s + %s (flagged as lazy, config has not changed since last sync)", d.opts.Path, contents.Path)
 			lockConfig.Contents = append(lockConfig.Contents, oldLockContents)
+			// copy previously fetched contents to staging dir
+			err = dircopy.Copy(filepath.Join(d.opts.Path, contents.Path), stagingDstPath)
+			if err != nil {
+				return lockConfig, fmt.Errorf("Copying existing content to staging '%s': %s", d.opts.Path, err)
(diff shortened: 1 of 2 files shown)


pkg/vendir/directory/directory.go
```diff
@@ -89,6 +89,11 @@ func (d *Directory) Sync(syncOpts SyncOpts) (ctlconf.LockDirectory, error) {
 		if skipFetching {
```
prembhaskal, line 89: A thought, Should we rename this variable to "fetchFromCache" because we are fetching from previously cached data?

fritzduchardt: Hi @prembhaskal, I am open to renaming the flag, but it should not contain the word "cache", since we copy from the already fetched final content location and not a cache, i.e. a separate copy that is kept to increase performance. "Fetching" in this context means for me "to pull from the source location defined in the vendir.yml".

prembhaskal: hmm fine. I think for current review it is fine. But we can think of changing name as a refactoring PR.

prembhaskal:
looks good for me. @joaopapereira  you can have a look at it when you get time.

joaopapereira (approved):
LGTM

PR #349, opened 2024-01-05
PrivateKey with or without extra char as newline will be accepted

### LINKED ISSUE #350: Vendir is failing to fetch content if private-key does not have newline at the end
**What steps did you take:**
**Steps to reproduce the issue:**
-> Copy the content from file ~/.ssh/id_rsa (assuming this is the private key file on your system)
-> encode the key by running cmd: `pbpaste | base64`
-> create vendir.yml
```
apiVersion: v1
kind: Secret
metadata:
  name: http-auth
data:
  ssh-privatekey:  <paste encoded value of private-key>
---
kind: Config
directories:
- path: vendor
  contents:
  - path: dump
    git:
      url: git@github.com:carvel-dev/vendir.git
      ref: origin/develop
      secretRef:
        name: http-auth
```
-> run vendir sync and you will see the error:
```
Load key "/Users/ktanushree/code_work/test-repo/.vendir-tmp/incoming/git-auth/private-key": invalid format
  git@github.com: Permission denied (publickey).
  fatal: Could not read from remote repository.
  
  Please make sure you have the correct access rights
  and the repository exists.

vendir: Error: Syncing directory 'vendor':
  Syncing directory 'dump' with git contents:
    Fetching git repository:
      Git [fetch origin main]: exit status 128 (stderr: Load key "/Users/ktanushree/code_work/test-repo/.vendir-tmp/incoming/git-auth/private-key": invalid format
git@github.com: Permission denied (publickey).
fatal: Could not read from remote repository.

Please make sure you have the correct access rights
and the repository exists.
)
```
Note: pbpaste does not add new line but if you will use cat instead of pbpaste you will see newline will get added to your private key. You can try same steps just encode your private key by running command: `cat ~/.ssh/id_rsa | base64`. This will work fine.

**What happened:**

**What did you expect:**
Vedir should work fine with/without newline in privatekey.

Git actually does not work if privatekey does not have newline at the end. So add newline when vendir is writing privateKey in file. Private key with multiple newline also accepted by git.

**Environment:**

- vendir version (execute `vendir --version`):

---
Vote on this request

kumaritanushree:
fix https://github.com/carvel-dev/vendir/issues/350

### DIFF
--- a/pkg/vendir/fetch/git/git.go
+++ b/pkg/vendir/fetch/git/git.go
@@ -93,7 +93,8 @@ func (t *Git) fetch(dstPath string, tempArea ctlfetch.TempArea) error {
 		if authOpts.PrivateKey != nil {
 			path := filepath.Join(authDir, "private-key")
 
-			err = os.WriteFile(path, []byte(*authOpts.PrivateKey), 0600)
+			// Ensure the private key ends with a newline character, as git requires it to work. (https://github.com/carvel-dev/vendir/issues/350)
+			err = os.WriteFile(path, []byte(*authOpts.PrivateKey+"\n"), 0600)
 			if err != nil {
 				return fmt.Errorf("Writing private key: %s", err)
 			}


100mik (approved):
Verified that these changes work! :GTM 🚀

joaopapereira:
Is there any kind of test that we can do to ensure that we do not break this in the future? If not we should add a comment to that line to explain why that new line is there

rohitagg2020:
LGTM

praveenrewar (commented):
+1 to what Joao said.
We already seem to be adding a newline while adding gitCreds (on line 141), should we check if there are other places where a newline needs to be added (maybe as part of a separate issue)?

kumaritanushree:
> Is there any kind of test that we can do to ensure that we do not break this in the future? If not we should add a comment to that line to explain why that new line is there

Added comment as of now as to add e2e test we will required to maintain dummy private repo and private key to access that. I think comment should be fine.

pkg/vendir/fetch/git/git.go
```diff
@@ -93,7 +93,9 @@ func (t *Git) fetch(dstPath string, tempArea ctlfetch.TempArea) error {
 		if authOpts.PrivateKey != nil {
 			path := filepath.Join(authDir, "private-key")
 
-			err = os.WriteFile(path, []byte(*authOpts.PrivateKey), 0600)
+			// Adding newlinne as git does not support private-key without newline as last character.
+			// Now vendir will work with both with/without newline in private-key.
+			err = os.WriteFile(path, []byte(*authOpts.PrivateKey+"\n"), 0600)
```
Zebradil, line 98: At first, I noticed a typo (`newlinne`), but decided to expand the suggestion. How about this?

```suggestion
			// Ensure the private key ends with a newline character, as git requires it to work.
			err = os.WriteFile(path, []byte(*authOpts.PrivateKey+"\n"), 0600)
```

pkg/vendir/fetch/git/git.go
```diff
@@ -93,7 +93,9 @@ func (t *Git) fetch(dstPath string, tempArea ctlfetch.TempArea) error {
 		if authOpts.PrivateKey != nil {
 			path := filepath.Join(authDir, "private-key")
 
-			err = os.WriteFile(path, []byte(*authOpts.PrivateKey), 0600)
+			// Adding newlinne as git does not support private-key without newline as last character.
+			// Now vendir will work with both with/without newline in private-key.
```
Zebradil: At first, I noticed a typo (`newlinne`), but decided to expand the suggestion. How about this?

```suggestion
			// Ensure the private key ends with a newline character, as git requires it to work.
```

kumaritanushree: updated

joaopapereira (approved):
Making a test for this would have been great but we would have needed to create private repositories and in some way share the keys which is not great and that would be an extra issue for local development. Also thinking a little bit more we could try to do some sort of unit tests for this but we might need to do a major refactor of this struct, which for now I think we should hold of.
Due to this I am ok with the comment added, and the PR is ready for merge when tests are green

PR #362, opened 2024-02-14
Force basic auth via flag

### LINKED ISSUE #361: Support forcing basic authentication for git
Using older versions of Team Foundation Server (now renamed to Azure DevOps Server, no relation to Azure) require git *basic* authentication, which requires a form of tricking the `git` cli that vendir uses in order to properly authenticate.  Specifically, [Use a PAT](https://learn.microsoft.com/en-us/azure/devops/organizations/accounts/use-personal-access-tokens-to-authenticate?view=azure-devops&tabs=Linux#use-a-pat) is asking for `-c http.extraHeader="Authorization: Basic ${B64_PAT}"` to be passed to the `git clone` step for cloning a repository.

If vendir detects the presence of an environment variable, like `VENDIR_GIT_ARGS` or something similar then those are added to the git commands that are executed.

Alternatively, and probably more robust, would be adding this directly into the vendir sync, although I'm unsure of how to pass credentials through that process then.  Also, since my ultimate solution needs to work in kapp-controller, that is going to require updates to the App CR spec to support as well.

I am unsure of how else to trick git into properly using basic authentication, it looks like Microsoft doesn't know either as that link is their official documentation.

---
Vote on this request

joaopapereira:
Refactor the code to allow for more granular tests

Fixes #361

### DIFF
--- a/pkg/vendir/config/directory.go
+++ b/pkg/vendir/config/directory.go
@@ -72,6 +72,7 @@ type DirectoryContentsGit struct {
 	DangerousSkipTLSVerify bool `json:"dangerousSkipTLSVerify,omitempty"`
 	SkipInitSubmodules     bool `json:"skipInitSubmodules,omitempty"`
 	Depth                  int  `json:"depth,omitempty"`
+	ForceHTTPBasicAuth     bool `json:"forceHTTPBasicAuth,omitempty"`
 }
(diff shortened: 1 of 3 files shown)


kumaritanushree (approved):
LGTM

PR #371, opened 2024-03-29
Introduce possibility for custom HTTP Headers in HTTP source

### LINKED ISSUE #87: Possibility to add custom headers for HTTP downloads
Some websites request to add specific HTTP headers, like for example API Token to access the files.
Currently it seems not possible to add those headers in the yaml file.
Example: wget -O <filename> --header="Authorization: Token *************" <url>

Add a new field in the vendir.yaml configuration file for each header.

Thank you :)

---
Vote on this request

christopherfriedrich:
This PR introduces the possibility to add custom HTTP Headers to requests for fetching HTTP sources and thereby fixes #87. 

The custom HTTP headers for the HTTP source can be provided via a new field `httpHeadersSecretRef` in the HTTP source spec. All key value pairs from that secret will be added as HTTP Header to the HTTP request to fetch the file from the HTTP source.

### DIFF
--- a/pkg/vendir/fetch/http/sync.go
+++ b/pkg/vendir/fetch/http/sync.go
@@ -86,6 +86,11 @@ func (t *Sync) downloadFile(dst io.Writer) error {
 		return fmt.Errorf("Adding auth to request: %s", err)
 	}
 
+	err = t.addHTTPHeaders(req)
+	if err != nil {
+		return fmt.Errorf("Adding HTTP Headers to request: %s", err)
+	}
+
 	resp, err := http.DefaultClient.Do(req)
 	if err != nil {
 		return fmt.Errorf("Initiating URL download: %s", err)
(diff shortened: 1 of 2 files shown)


pkg/vendir/fetch/http/sync.go
```diff
@@ -160,3 +165,17 @@ func (t *Sync) addAuth(req *http.Request) error {
...
+
+func (t *Sync) addHTTPHeaders(req *http.Request) error {
+	if t.opts.HTTPHeadersSecretRef != nil {
+		secret, err := t.refFetcher.GetSecret(t.opts.HTTPHeadersSecretRef.Name)
+		if err != nil {
+			return err
+		}
+
+		for name, value := range secret.Data {
+			req.Header.Set(name, string(value))
```
christopherfriedrich, line 177: **I've made a design decision here you may want to revisit**: This introduces the possibility, that the Basic Authentication headers, set via the `secretRef` field, may be overwritten, if there is a value for that Header provided in the secret for `httpHeadersSecretRef`. 

I could also rework this and prohibit the overwriting for the Basic Auth headers if you like. Nevertheless I'll open a PR to also document this properly in either case.

praveenrewar (commented):
Thank you so much for the PR @christopherfriedrich!!
I think the changes mostly looks good. @christopherfriedrich Do you think you would be able to add some tests for this?

kumaritanushree:
> Thank you so much for the PR @christopherfriedrich!! I think the changes mostly looks good. @christopherfriedrich Do you think you would be able to add some tests for this?

+1

kumaritanushree (commented):
> Thank you so much for the PR @christopherfriedrich!!
I think the changes mostly looks good. @christopherfriedrich Do you think you would be able to add some tests for this?

+1

pkg/vendir/fetch/http/sync.go
joaopapereira: I think we should prevent at least the Auth header from being updated.

PR #372, opened 2024-04-02
hg: Add a cache for mercurial repositories.

cdevienne:
hg: Add a cache for mercurial repositories.

The idea is to save the whole untouched clone (with no checkout) in the cache.

If already present, the fetch (a pull actually for hg) is done directly in the cache, and is
faster (except on very small repos) because only new changesets are
transfered.

If the ref is a changeset id (not a tag, branch, topic or bookmark), and
the changeset is already known in the cached clone, no pull is done
which avoid any network exchange.

Then we copy the cached entry and do the checkout.

The same feature for git will be proposed very soon, so any feedback on this one is welcome if it helps doing things better for git!


pkg/vendir/fetch/hg/hg.go
```diff
@@ -21,12 +21,28 @@ type Hg struct {
...
+	}
+	return &t, nil
+}
 
-	return &Hg{opts, infoLog, refFetcher}
+// CacheID returns a cache id for the repository
+// It doesn't include the ref because we want to reuse a cache when only the ref
+// is changed
+// Basically we combine all data used to write the hgrc file
+func (t *Hg) CacheID() string {
```
kumaritanushree: Changes looks good to me but as being not an expert in hg having low confidence on what all data this cacheID will contain. 

Here we are saying  it won't include the ref (from my understanding could be tag, branch and commit id) and in case of commit id, it will change only  if there is content change. And in this case we will not want to use from cache. 
I am bit confuse how we are saying that the target what we want is same as what we get from cache if we are not storing ref in cache?

Or we are considering this for only tag and branch?

pkg/vendir/fetch/hg/hg.go
```diff
@@ -35,27 +51,60 @@ type HgInfo struct {
 	ChangeSetTitle string
 }
 
-func (t *Hg) Retrieve(dstPath string, tempArea ctlfetch.TempArea) (HgInfo, error) {
-	if len(t.opts.URL) == 0 {
-		return HgInfo{}, fmt.Errorf("Expected non-empty URL")
+// CloneHasTargetRef returns true if the given clone contains the target
+// ref, and this ref is a revision id (not a tag or a branch)
+func (t *Hg) CloneHasTargetRef(dstPath string) bool {
```
kumaritanushree: In cachID() we are saying we are not including ref but here we are matching content with revisionID. 

May be I am not able to understand it correctly.  @cdevienne can you please help me to understand it bit more?
Thanks

kumaritanushree:
@cdevienne not  sure how much effort will require but if possible can you please add test for this?

pkg/vendir/fetch/hg/hg.go
cdevienne: When we have an entry in the cache, we know it is the right repository.
From there we need to make sure it already contains the right changeset, which can be done by simply fetching changes. But if we tracked by changeset ID, we can test for its presence directly, avoiding a network roundtrip. But we cannot do it for tags and branches as they may change each time we fetch changes.

pkg/vendir/fetch/hg/hg.go
cdevienne: The ref can be a changeset id, a tag, a branch, a bookmark, a topic: anything that can designate a mercurial changeset.

No checkout is done on the cached repository, so it is not pointing any particular changeset (in git terms, it has no HEAD).
The goal is to reuse this repository as a base whenever we sync any changeset from the same source URL.

That said the cache ID I wrote had 2 issues:
- it did not contain the URL
- it leaked credentials

I fixed those two issues in the next changeset.

cdevienne:
@kumaritanushree I will write some tests. Do you need more precisions on the CacheID logic I implemented?

pkg/vendir/fetch/hg/hg.go
kumaritanushree: > But if we tracked by changeset ID, we can test for its presence directly, avoiding a network roundtrip

Are you saying that changesetID( commit ID in term of git) we are storing in cache to compare it later when we found the repo in cache?
If yes, then I found the contradictory statements when we are generating cacheID and when we are reading/matching details in CloneHasTargetRef()

I think once you will add e2e test might be more helpful to understand.

test/e2e/hg_test.go
```diff
@@ -0,0 +1,167 @@
...
+	env := BuildEnv(t)
+	logger := Logger{}
+	vendir := Vendir{t, env.BinaryPath, logger}
+
+	_ = Vendir{t, env.BinaryPath, logger}
+
+	hgSrcPath, err := os.MkdirTemp("", "vendir-e2e-hg-cache")
+	if err != nil {
+		t.Fatal(err)
+	}
```
joaopapereira:
```suggestion
	require.NoError(t, err)
```

test/e2e/hg_test.go
```diff
@@ -0,0 +1,167 @@
...
+	require.NoError(t, err)
+	defer os.RemoveAll(cachePath)
+
+	dstPath, err := os.MkdirTemp("", "vendir-e2e-hg-cache-dst")
+	require.NoError(t, err)
+	defer os.RemoveAll(dstPath)
+
+	var stdout bytes.Buffer
+	stdoutDec := json.NewDecoder(&stdout)
+	//defer os.RemoveAll(dstPath)
```
joaopapereira:
```suggestion
```

pkg/vendir/fetch/hg/sync.go
```diff
@@ -44,11 +46,37 @@ func (d Sync) Sync(dstPath string, tempArea ctlfetch.TempArea) (ctlconf.LockDire
...
+		// fetch from cachedDir
+		if err := d.cache.CopyFrom("hg", hg.CacheID(), incomingTmpPath); err != nil {
+			return hgLockConf, fmt.Errorf("Extracting cached hg clone: %w", err)
+		}
+	} else {
+		// fetch in the target directory, and save it to cache
+		if err := hg.Clone(incomingTmpPath); err != nil {
+			return hgLockConf, fmt.Errorf("Cloning hg repository: %w", err)
+		}
+		if err := d.cache.Save("hg", hg.CacheID(), incomingTmpPath); err != nil {
```
joaopapereira: For OCI images we are making a decision to cache only when the image that vendir is retrieving is immutable (addressed by SHA). Does it make sense for us to also do the same thing here? If the user provides a ref that is not a SHA maybe we should not cache the repository.

pkg/vendir/fetch/hg/hg.go
```diff
@@ -134,6 +194,7 @@ hgauth.password = %s
 			}
 
 			sshCmd = append(sshCmd, "-o", "StrictHostKeyChecking=yes", "-o", "UserKnownHostsFile="+path)
+			cacheID += "known-hosts=" + *authOpts.KnownHosts + "|"
```
joaopapereira: What are we trying to prevent with the KownHosts inclusion in the CacheID?

pkg/vendir/fetch/hg/hg.go
```diff
@@ -147,39 +208,39 @@ hgauth.password = %s
 		if err != nil {
 			return fmt.Errorf("Writing %s: %s", hgRcPath, err)
 		}
-		env = append(env, "HGRCPATH="+hgRcPath)
+		t.env = append(t.env, "HGRCPATH="+hgRcPath)
+		cacheID += hgRc
```
joaopapereira: Since we are creating a temp folder to store auth, isn't that going to invalidate the cache every time we run the command and use an SSH Private Key?

joaopapereira (changes requested):
Added some questions and left some comments.

pkg/vendir/fetch/hg/sync.go
cdevienne: The goal here is to store as much history as we can, so further pulls are much quicker, not to store an exact version.

pkg/vendir/fetch/hg/hg.go
cdevienne: Nothing really, I realize the only key we need is the repository URL. Will do that.

pkg/vendir/fetch/hg/hg.go
cdevienne: hgRc doesn't contain the temp folder path, so no. But as I said in the above thread, we only need the URL.

pkg/vendir/fetch/hg/sync.go
joaopapereira: The only issue I am concerned about here is that we are not clearing the caching, nor do we have an idea of how large our cache is currently. Due to this, if we save everything every time, it will cut down fetch times but is unbounded, which means it will grow in size forever.
Maybe we will have to rethink the caching process if this is a feature that we really want.
@Zebradil do you have any suggestion here?

joaopapereira (changes requested):
Do you mind fixing the rebase so that we do not have all the other changes in this PR?

pkg/vendir/fetch/hg/sync.go
Zebradil: The caching topic is pretty much in line with what we're currently working on in myks (in mykso/myks#274, it is almost done and will be merged in the next few days, hopefully). The idea was to test the approach there and then to made a proposal to vendir.

And I was thinking about the cache cleanup approach just yesterday.

That being said, I'll take a look at this PR over the weekend and get back with some particular ideas.

cc @fritzduchardt

cdevienne: We can do the following:
- copy the repo from the cache
- pull into the copy
- save the copy to the cache

Doing so add "only" a full copy of the repo, which is still way cheaper than a full clone. And it allows the cache system to reevaluate the entry size, solving uncontrolled growth of the cache.

I think the added safety (no unbounded growth) is worth the drawback, I will update my PR accordingly.

Zebradil:
Apologies for the delay with the review.

As I haven't been using mercurial for more than a decade, it's hard to think through the hg-related logic. However, I still have a couple of notes.

@cdevienne I omit some not important things, but this one I would like to address before this gets merged: can we not export the new functions if they are not used in other packages? Otherwise, they'll become a part of the public API and it'll be harder changing them in the future.

@joaopapereira Regarding the immutability and cache size questions. I think that `vendir` can give a little more control over this to a user.

With references mutable by design, we can't decide with a high confidence whether it is safe to cache or not. Users, on the other hand, have more knowledge about artifacts they're dealing with and can explicitly allow to cache some of them with mutable references. It is totally fine as far as it is an opt-in.

The cache size restriction in the current implementation (if I read it correctly) is not ideal, as it is silently refuses to save new data, which is probably important, into the cache instead of ejecting less important data (e.g. oldest entries or least used). If I were to use `vendir` cache, I'd just set its size to some large number hoping that the cache size will remain smaller than that.

<hr>

I would merge this PR (after unexporting the new functions) and plan improvements to the caching layer, namely to implement a clean up logic that removes less important entries. I'd also allow users to explicitly request unlimited cache size, instead of the implicit `1Mb` limit, but we can discuss these details in a separate issue, I guess.

cdevienne:
I un-exported the "Hg" type completely, as it has no reasonable use outside this package.

Zebradil:
@joaopapereira should I be able to approve workflow runs?

praveenrewar:
@Zebradil Are you not able to approve? Let me check the permissions in that case. For now I am approving.

Zebradil:
@praveenrewar No, I didn't have the button.

Zebradil:
@cdevienne thank you for the adjustments. Now we can see that the `localClone` function is unused. Could you remove it? Or was it supposed to be used at some point?

praveenrewar:
> No, I didn't have the button.

@Zebradil Please check now. I am still figuring out some of the permissions stuff, so please excuse me if it still doesn't work 😅

Zebradil:
I can't test if it works now, as there are no pipelines to approve :)
Let's wait until we have another PR.

cdevienne:
> @cdevienne thank you for the adjustments. Now we can see that the `localClone` function is unused. Could you remove it? Or was it supposed to be used at some point?

my bad I forgot to run golangci-lint. I cleaned up the code.

Zebradil:
@cdevienne sorry, I didn't notice earlier that with the new functions some existing structs (`Hg`, `HgInfo`, ...) and functions (`NewHg`, ...) became unexported too. This could be a breaking change for someone who might use them in their code. Could you revert this change to the said structs and functions?

Zebradil:
@praveenrewar I don't see the button to approve workflow runs.

image

cdevienne:
> @cdevienne sorry, I didn't notice earlier that with the new functions some existing structs (`Hg`, `HgInfo`, ...) and functions (`NewHg`, ...) became unexported too. This could be a breaking change for someone who might use them in their code. Could you revert this change to the said structs and functions?

It is a breaking change indeed, but I would be surprised it breaks anything. I don't see a point in keeping these public if it doesn't break any code. If it does I would happily revert the change.

cdevienne:
Precising my point: I think the internal of the fetchers are not a part of the contract vendir has with its users. So it should not have been public in the first place.

Zebradil:
> I think the internal of the fetchers are not a part of the contract vendir has with its users.

I doubt that this code is used outside of vendir. However, it is located under `pkg` which, if I understand correctly, makes it "public". If it were under `internal` I wouldn't bother much.

> I don't see a point in keeping these public if it doesn't break any code.

Yes, but how do we know if it doesn't break someone else's code?

@joaopapereira could you give a suggestion here?

Removing symbols from public API is a breaking change according to the semantic versioning spec and we should implement it carefully with adequate communication. But vendir's major version is still `0` which gives us some freedom to break the public API. **I personally would commit this (breaking) change** and see who reacts on this and what are their use cases. Also, ytt, for example, discourages using it as a module: https://github.com/carvel-dev/ytt/blob/develop/examples/integrating-with-ytt/apis.md#as-a-go-module. I think for vendir these considerations make sense too.

joaopapereira:
As a generic rule of thumb in Carvel we tend to:
- Minimize the breaking changes, our tools are still all in the major 0 version, which gives us some leeway to do breaking changes. Nevertheless, we should aim to minimize it
- We try to keep consistency whenever possible

I like the idea of exposing only the sync, but we should standardize what this sync interface should have before we do that. This being said, I would recommend we keep exported Structs exported, and in the future, if we want to invest time in getting an interface going that would support all the fetching types, we can create the interface and make all things private that we want to make private.

In terms of API consumption for the majority of the tools, we expect people to use the Cobra commands interface (not the best experience) until we are in a position where we have a supported API for each tool like we currently have in [imgpkg](https://github.com/carvel-dev/imgpkg/tree/develop/pkg/imgpkg/v1).

joaopapereira:
> @praveenrewar I don't see the button to approve workflow runs.
> 
> image

Only folks with write access can see that button, and write access comes with the Approver Role.

cdevienne:
Should I make the structs public again ? What about their functions ?
Or can it be merged as it is ?

joaopapereira:
Let us make them public again and we will think in the future on a way to make a good interface API for this

cdevienne:
ok doing it. Be aware that the 'Retrieve' function does not exist anymore, and that the Hg struct has no exposed function aside from a constructor and a Close function.

cdevienne:
Just to be clear: the Hg public interface is changing, an the new one is useless as it is. Which is why I unexposed it completely.
Pushing the change you asked for anyway, let me know if you need any other change.

cdevienne:
Hi @joaopapereira, any chance we merge this PR soon ?

joaopapereira (approved):
LGTM

joaopapereira:
@Zebradil can you give it a last review please and if you are ok with it I will merge the PR

Zebradil (approved):
LGTM

cdevienne:
Thanks!

PR #380, opened 2024-05-10
git: add a cache based on bundles

cdevienne:
The cache strategy is simpler than the one I did for mercurial, and a little less efficient (it is slower).
But it is solid, and avoids a cumbersome refactoring of the fetcher.

### DIFF
--- a/pkg/vendir/fetch/git/sync.go
+++ b/pkg/vendir/fetch/git/sync.go
@@ -51,14 +51,7 @@ func (d Sync) Sync(dstPath string, tempArea ctlfetch.TempArea, cache ctlcache.Ca
 
 	defer os.RemoveAll(incomingTmpPath)
 
-	sum := sha256.New()
-	if _, err := sum.Write([]byte(d.opts.URL)); err != nil {
-		if err != nil {
-			return gitLockConf, err
-		}
-	}
-
-	cacheID := string(sum.Sum(nil))
+	cacheID := fmt.Sprintf("%x", sha256.Sum256([]byte(d.opts.URL)))
 
 	git := NewGit(d.opts, d.log, d.refFetcher)
 
@@ -66,7 +59,6 @@ func (d Sync) Sync(dstPath string, tempArea ctlfetch.TempArea, cache ctlcache.Ca
 	if cacheEntry, hasCache := cache.Has(gitCacheType, cacheID); hasCache {
 		bundle = filepath.Join(cacheEntry, "bundle")
 	}
-	fmt.Println("We got cache?", cacheID, bundle)
 
 	info, err := git.Retrieve(incomingTmpPath, tempArea, bundle)
 	if err != nil {
@@ -77,7 +69,7 @@ func (d Sync) Sync(dstPath string, tempArea ctlfetch.TempArea, cache ctlcache.Ca
 	gitLockConf.Tags = info.Tags
 	gitLockConf.CommitTitle = d.singleLineCommitTitle(info.CommitTitle)
 
-	if os.Getenv("VENDIR_CACHE_DIR") != "" {
+	if _, ok := cache.(*ctlcache.NoCache); !ok {
 		// attempt to save a bundle to the cache
 		bundleDir, err := tempArea.NewTempDir("bundleCache")
 		if err != nil {
--- a/pkg/vendir/fetch/git/sync.go
+++ b/pkg/vendir/fetch/git/sync.go
@@ -4,14 +4,19 @@
 package git
 
 import (
+	"crypto/sha256"
 	"fmt"
 	"io"
 	"os"
+	"path/filepath"
 	"strings"
 
 	ctlconf "carvel.dev/vendir/pkg/vendir/config"
 	ctlfetch "carvel.dev/vendir/pkg/vendir/fetch"
+	ctlcache "carvel.dev/vendir/pkg/vendir/fetch/cache"
 )
+
+const gitCacheType = "git-bundle"
 
 type Sync struct {
 	opts       ctlconf.DirectoryContentsGit
@@ -36,7 +41,7 @@ func (d Sync) Desc() string {
 	return fmt.Sprintf("%s@%s", d.opts.URL, ref)
 }
 
-func (d Sync) Sync(dstPath string, tempArea ctlfetch.TempArea) (ctlconf.LockDirectoryContentsGit, error) {
+func (d Sync) Sync(dstPath string, tempArea ctlfetch.TempArea, cache ctlcache.Cache) (ctlconf.LockDirectoryContentsGit, error) {
 	gitLockConf := ctlconf.LockDirectoryContentsGit{}
 
 	incomingTmpPath, err := tempArea.NewTempDir("git")
@@ -46,9 +51,24 @@ func (d Sync) Sync(dstPath string, tempArea ctlfetch.TempArea) (ctlconf.LockDire
 
 	defer os.RemoveAll(incomingTmpPath)
 
+	sum := sha256.New()
+	if _, err := sum.Write([]byte(d.opts.URL)); err != nil {
+		if err != nil {
+			return gitLockConf, err
+		}
+	}
+
+	cacheID := string(sum.Sum(nil))
+
 	git := NewGit(d.opts, d.log, d.refFetcher)
 
-	info, err := git.Retrieve(incomingTmpPath, tempArea)
+	var bundle string
+	if cacheEntry, hasCache := cache.Has(gitCacheType, cacheID); hasCache {
+		bundle = filepath.Join(cacheEntry, "bundle")
+	}
+	fmt.Println("We got cache?", cacheID, bundle)
+
+	info, err := git.Retrieve(incomingTmpPath, tempArea, bundle)
 	if err != nil {
 		return gitLockConf, fmt.Errorf("Fetching git repository: %s", err)
 	}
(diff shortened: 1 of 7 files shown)


cdevienne:
Hello @Zebradil,
Any chance this PR gets reviewed soon?
Thanks!

Zebradil:
Hey @cdevienne, sorry for the delay. As I'm currently not allowed to approve workflow runs on my own, I was waiting for the checks to be completed. But apparently, checks are not emitting any notifications upon completion, so this PR fell out of my sight. I'll check it out today or tomorrow.

pkg/vendir/fetch/git/sync.go
```diff
@@ -46,9 +51,24 @@ func (d Sync) Sync(dstPath string, tempArea ctlfetch.TempArea) (ctlconf.LockDire
...
 	defer os.RemoveAll(incomingTmpPath)
 
+	sum := sha256.New()
+	if _, err := sum.Write([]byte(d.opts.URL)); err != nil {
+		if err != nil {
+			return gitLockConf, err
+		}
+	}
+
+	cacheID := string(sum.Sum(nil))
```
Zebradil: The nested `if err != nil` is redundant.

What about simplifying the code like this:

```suggestion
	cacheID := fmt.Sprintf("%x", sha256.Sum256([]byte(d.opts.URL)))
```

pkg/vendir/fetch/git/sync.go
```diff
@@ -46,9 +51,24 @@ func (d Sync) Sync(dstPath string, tempArea ctlfetch.TempArea) (ctlconf.LockDire
...
+
 	git := NewGit(d.opts, d.log, d.refFetcher)
 
-	info, err := git.Retrieve(incomingTmpPath, tempArea)
+	var bundle string
+	if cacheEntry, hasCache := cache.Has(gitCacheType, cacheID); hasCache {
+		bundle = filepath.Join(cacheEntry, "bundle")
+	}
+	fmt.Println("We got cache?", cacheID, bundle)
```
Zebradil: Leftover from debugging?

```suggestion
```

pkg/vendir/fetch/git/sync.go
```diff
@@ -57,6 +77,35 @@ func (d Sync) Sync(dstPath string, tempArea ctlfetch.TempArea) (ctlconf.LockDire
 	gitLockConf.Tags = info.Tags
 	gitLockConf.CommitTitle = d.singleLineCommitTitle(info.CommitTitle)
 
+	if os.Getenv("VENDIR_CACHE_DIR") != "" {
```
Zebradil: This check is already done here: https://github.com/carvel-dev/vendir/blob/f59017e59730f530e866ae749a4da107db90eb81/pkg/vendir/fetch/cache/cache.go#L35

Here you can check if the `cache` argument is `NoCache` using something like this:

```go
if _, ok := cache.(*NoCache); !ok {
// ...
}
```

pkg/vendir/fetch/git/sync.go
```diff
@@ -36,7 +41,7 @@ func (d Sync) Desc() string {
 	return fmt.Sprintf("%s@%s", d.opts.URL, ref)
 }
 
-func (d Sync) Sync(dstPath string, tempArea ctlfetch.TempArea) (ctlconf.LockDirectoryContentsGit, error) {
+func (d Sync) Sync(dstPath string, tempArea ctlfetch.TempArea, cache ctlcache.Cache) (ctlconf.LockDirectoryContentsGit, error) {
```
Zebradil: Instead of passing `cache` as an argument to this function, I'd use the approach of the Imgpkg syncer: https://github.com/carvel-dev/vendir/blob/f59017e59730f530e866ae749a4da107db90eb81/pkg/vendir/fetch/image/sync.go#L15-L27

It carries the cache object within the `Sync` struct, making it available to all its functions. Then, there might a function `getBundleName` for getting the bundle name (if exists) from the cache. `getBundleName` function can be used in the `fetch` function directly.

These two measures will allow to keep signatures of the functions (except the `git.NewSync`) unchanged.

Zebradil (changes requested):
First of all, many thanks for bringing this topic into attention and filing this PR!

The idea of using bundle seems okay. However, I have a couple of questions about it:

- Did you consider maintaining a working copy in the cache instead of a bundle? (Earlier we had a discussion about the cache size. I see that the bundle approach allows to keep the cache size under control, which is hard when using the clone approach. But are there any other considerations?)
- How does it work when history of a remote git repo gets rewritten (with `git push --force`, for example)?

Apart from that I left a few code-related comments. These suggestions should make the code a little more straightforward and easier to change in the future.

@joaopapereira Another thing I noticed is that now the `vendir sync` output can be quite mouthful, as it prints all refs in the repo being synced two times (in addition to the output of `git fetch origin` we can see in the main version of vendir).

*UPD:* This is not an issue of this PR, it is how git command runner works in vendir: it prints full output of a command. I'm not sure how to go about that. Maybe we should make verbosity level configurable in the future.

pkg/vendir/fetch/git/sync.go
cdevienne: nice !

pkg/vendir/fetch/git/sync.go
cdevienne: I move the cache to NewSync.

cdevienne: about getBundleName, if we want to avoid an extra argument to git.Retrieve, we would need to put it on the Git struct. And I have a feeling it is not the job of the Git type.

cdevienne:
>
> * Did you consider maintaining a working copy in the cache instead of a bundle? (Earlier we had a discussion about the cache size. I see that the bundle approach allows to keep the cache size under control, which is hard when using the clone approach. But are there any other considerations?)
>

Yes I did, but it requires a bigger refactoring of the fetch code, and a deeper knowledge of git.
The approach I chose is very safe, and has no corner case I can think of.

> 
> * How does it work when history of a remote git repo gets rewritten (with `git push --force`, for example)?

Then the git fetch would get all the new commits and cache a new updated bundle.

> _UPD:_ This is not an issue of this PR, it is how git command runner works in vendir: it prints full output of a command. I'm not sure how to go about that. Maybe we should make verbosity level configurable in the future.

Yes that would be nice.

Zebradil (approved):
LGTM,

except the fact that the public interface is changed (new argument to `Retrieve`). However, to me it's a minor issue, as I don't think that it'll bring any problems to users.

@joaopapereira could you take a look as well?

joaopapereira:
> @joaopapereira Another thing I noticed is that now the vendir sync output can be quite mouthful, as it prints all refs in the repo being synced two times (in addition to the output of git fetch origin we can see in the main version of vendir).
> 
> UPD: This is not an issue of this PR, it is how git command runner works in vendir: it prints full output of a command. I'm not sure how to go about that. Maybe we should make verbosity level configurable in the future.

do you want to create an issue with that?

joaopapereira:
I will try to review it next week

joaopapereira:
The PR looks good, is there any way we can have some tests around the caching?

cdevienne:
I added a basic test to prove it unbundles from the cache.

test/e2e/git_test.go
```diff
@@ -133,6 +136,110 @@ directories:
...
+			Dir:          dstPath,
+			StdinReader:  yamlConfig("signed-trusted-tag"),
+			StdoutWriter: &stdout,
+			Env: []string{
+				"VENDIR_CACHE_DIR=" + tmpDir,
+				"VENDIR_CACHE_MAX_SIZE=10M",
+			},
+		})
+
+	var out hgVendirOutput
```
joaopapereira: If we are going to use it here, maybe we should change hgVendirOutput to vendirOutput, because this threw me off when I was ready the test

test/e2e/git_test.go
```diff
@@ -133,6 +136,110 @@ directories:
...
+
+	require.NoError(t, stdoutDec.Decode(&out))
+
+	var unbundled bool
+	for _, l := range out.Lines {
+		if strings.Contains(l, "unbundle") {
+			unbundled = true
+		}
+	}
+	assert.True(t, unbundled, "git did not use the cached bundle")
```
joaopapereira:
```suggestion
	require.True(t, unbundled, "git did not use the cached bundle")
```

test/e2e/git_test.go
```diff
@@ -133,6 +136,110 @@ directories:
...
+	vendir := Vendir{t, env.BinaryPath, logger}
+
+	tmpDir, err := os.MkdirTemp("", "vendir-e2e-git-verify-signed-git-repo")
+	require.NoError(t, err)
+	defer os.RemoveAll(tmpDir)
+
+	gitSrcPath, err := os.MkdirTemp("", "vendir-e2e-git-verify-signed-git-repo")
+	if err != nil {
+		t.Fatal(err)
+	}
```
joaopapereira:
```suggestion
	require.NoError(t, err)
```

test/e2e/git_test.go
```diff
@@ -133,6 +136,110 @@ directories:
...
+	defer os.RemoveAll(gitSrcPath)
+
+	if out, err := exec.Command("tar", "xzvf", "assets/git-repo-signed/asset.tgz", "-C", gitSrcPath).CombinedOutput(); err != nil {
+		t.Fatalf("Unpacking git-repo-signed asset: %s (output: '%s')", err, out)
+	}
+
+	dstPath, err := os.MkdirTemp("", "vendir-e2e-git-verify-signed-dst")
+	if err != nil {
+		t.Fatal(err)
+	}
```
joaopapereira:
```suggestion
	require.NoError(t, err)
```

test/e2e/git_test.go
```diff
@@ -133,6 +136,110 @@ directories:
...
+
+	gitSrcPath, err := os.MkdirTemp("", "vendir-e2e-git-verify-signed-git-repo")
+	if err != nil {
+		t.Fatal(err)
+	}
+	defer os.RemoveAll(gitSrcPath)
+
+	if out, err := exec.Command("tar", "xzvf", "assets/git-repo-signed/asset.tgz", "-C", gitSrcPath).CombinedOutput(); err != nil {
+		t.Fatalf("Unpacking git-repo-signed asset: %s (output: '%s')", err, out)
+	}
```
joaopapereira:
```suggestion
	out, err := exec.Command("tar", "xzvf", "assets/git-repo-signed/asset.tgz", "-C", gitSrcPath).CombinedOutput()
	require.NoError(t, err)
```

Zebradil:
> > @joaopapereira Another thing I noticed is that now the vendir sync output can be quite mouthful, as it prints all refs in the repo being synced two times (in addition to the output of git fetch origin we can see in the main version of vendir).
> > UPD: This is not an issue of this PR, it is how git command runner works in vendir: it prints full output of a command. I'm not sure how to go about that. Maybe we should make verbosity level configurable in the future.
> 
> do you want to create an issue with that?

Good idea. #387

joaopapereira (changes requested):
Added some comments to the test, just stylistic. When that is fixed I think we can merge the PR.

test/e2e/git_test.go
cdevienne: indeed forgot about it

test/e2e/git_test.go
cdevienne: done, but kept the custom message

joaopapereira (approved):
LGTM

PR #401, opened 2024-10-09
Stricter caching for imgpkg bundles

### LINKED ISSUE #399: Account for nested `imgpkg` bundles in the caching logic
`imgpkg` bundles might contain references to other imgpkg bundles (as in the case of `PackageRepository` bundles).  Purely SHA based caching might lead to nested content being pulled from an older registry when a migration is attempted.

Relevant bug in kapp-controller: carvel-dev/kapp-controller#1603

Instead of accounting for just the SHA as a key while caching content, the entire registry URL should be considered.

This would not be an issue in a kapp-controller upgrade scenario as an update would roll out new pods, dropping the old cache.

---
Vote on this request

100mik:
Use a fully-qualified reference as key for a cache rather than the digest.

Resolves #399 and eventually carvel-dev/kapp-controller#1603

### DIFF
--- a/pkg/vendir/fetch/image/imgpkg.go
+++ b/pkg/vendir/fetch/image/imgpkg.go
@@ -63,8 +63,8 @@ func (t *Imgpkg) FetchBundleRecursively(imageRef, destination string) (string, e
 		return "", err
 	}
 
-	if _, hit := t.cache.Has(ImgpkgBundleArtifactType, ref.Identifier()); hit {
-		return imageRef, t.cache.CopyFrom(ImgpkgBundleArtifactType, ref.Identifier(), destination)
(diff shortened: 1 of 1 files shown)


100mik:
Confirmed that we will still only be caching artifacts with digest references as `imgpkg` code decides wether or not it a bundle is cacheable, and that remains unchanged.

praveenrewar (commented):
It would be good to have a test added for this.

joaopapereira (changes requested):
THe change looks good, but I agree with @praveenrewar we should have a test to ensure the behavior is what we expect

joaopapereira:
@100mik any update on this?

joaopapereira:
@100mik do you think you can add some tests to ensure that we will not have regressions in the future?

himsngh:
Fix is merged via https://github.com/carvel-dev/vendir/pull/465

PR #427, opened 2025-10-13
fix(git): optimize fetch to selectively include tags and checkout exa…

### LINKED ISSUE #102: For faster sync, add option to only download a single branch
Currently, `vendir` takes long time to clone a repository. If I am after files from a specific branch it takes long time and it downloads unnecessary stuff I don't care about.

Add an option to
- specify cloning depth
- only clone single branch 

```yml
kind: Config
directories:
- path: docs/modules
  contents:
  - path: ROOT/partials
    git:
      url: https://github.com/openshift/openshift-docs
      ref: rhacs-docs
      single-branch: true
      depth: 1
    includePaths:
    - modules/*
    newRootPath: modules/
```

---
Vote on this request

fpytloun:
- Disabled unconditional addition of "remote.origin.tagOpt" config; only add if not fetching exact ref
- Introduced exactRef flag to detect when fetching a specific origin ref
- When fetching exact ref, checkout uses "FETCH_HEAD" instead of the ref name to avoid detached head advice
- Cleaned up commented-out code and improved fetch argument logic accordingly

Fixes: https://github.com/carvel-dev/vendir/issues/102

### DIFF
--- a/pkg/vendir/fetch/git/git.go
+++ b/pkg/vendir/fetch/git/git.go
@@ -166,17 +166,20 @@ func (t *Git) fetch(dstPath string, tempArea ctlfetch.TempArea, bundle string) e
 		}
 	}
 
-	argss = append(argss, []string{"config", "remote.origin.tagOpt", "--tags"})
-
 	if bundle != "" {
 		argss = append(argss, []string{"bundle", "unbundle", bundle})
 	}
 
+	exactRef := false
 	{
 		fetchArgs := []string{"fetch", "origin"}
 		if strings.HasPrefix(t.opts.Ref, "origin/") {
 			// only fetch the exact ref we're seeking
 			fetchArgs = append(fetchArgs, t.opts.Ref[7:])
+			exactRef = true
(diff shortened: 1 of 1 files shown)


joaopapereira (changes requested):
I am a little bit concerned with this change since we tried removing that same line before and failed for some cases like: https://github.com/carvel-dev/vendir/issues/263
Can you please confirm that a we will not have a regression here?

fpytloun:
> I am a little bit concerned with this change since we tried removing that same line before and failed for some cases like: #263 Can you please confirm that a we will not have a regression here?

I read that issue and assuming failure was this: https://github.com/carvel-dev/vendir/issues/263#issuecomment-1583253263

The trick is to simply checkout FETCH_HEAD instead of unresolved tag 🙂

joaopapereira (changes requested):
In terms of the solution looks good.  Can you please check why the tests are failing and if there's is something that can be done about the linting errors?

Issue #388, opened 2024-06-26
A 'dev' mode to preserve local changes in git/hg repositories


cdevienne:
**Describe the problem/challenge you have**

We use vendir in meta-projects, and work directly in the git/hg synced sub-repositories.
After doing some changes in them, we cannot do a "vendir sync" without loosing local changes (and lazy doesn't work if we pull some ref changes).
Also, after commit & pushing changes in the sub-repositories, we would like to automatically update the vendir configuration file with the current commit id/tag of each repo.

**Describe the solution you'd like**

We think vendir could benefit from a "dev" mode, which would be enabled by using a "dev-sync" command or a "--dev" switch on the sync command.

When turned on, vendir would change its sync behavior for mercurial and git (at least) in the following way:
- if the target directory already exists, keep it
- when using already existing target : 
  * check if non-committed changes exists, in which case the sync is aborted
  * fetch & checkout directly in the target directory

A new command "dev-baseline" (for example) would, for each vcs directory, determine what the current ref is:
- if current revision has a tag, the tag
- otherwise, the sha of the current revision
If the ref in the vendir file is not a branch (or a topic the mercurial case), and does not match the current ref determined above, the configuration is updated with the new current ref.

**Anything else you would like to add:**

The overall experience is inspired by the "hg-confman" mercurial extension that we intend to replace with vendir.

We are willing to work on an implementation for these features, and would like your opinion before diving into coding.

praveenrewar:
> We use vendir in meta-projects, and work directly in the git/hg synced sub-repositories.

@cdevienne Could you through some more light on this use case. vendir is used to pull config from a source. It seems like you also want to make changes in the same config and then be able to pull more changes from the source, which would make vendir work on two different source of truths.
Also, it would require vendir to do things like merging the config from source and local.
cc @joaopapereira

cdevienne:
One of our use cases is the following: we have repositories for odoo modules, that we sync in a top-level client project. When working on the client project, the developer updates some code in synced sub-repo(s), runs some tests locally and when satisfied commit changes in both the synced sub-repo(s) and the top-level repo.

When working a sub-repo, there is a period of time where the synced directory is not in-sync. So not really 2 sources of truth, just a gray area, until one those 2 events happen:

- vendir sync is called, and the reference expected by the vendir configuration is checked out
- the vendir configuration is updated to match the local reference

[...]

PR #436, opened 2025-12-29
Add "status" and "baseline" commands, add a "--safe" sync flag

cdevienne:
This PR implements my ideas for addressing #388.

1. The new "status" command gives a detailed status of the hg & git repositories (when they are complete).
2. The new "baseline" command updates the hg & git references in the vendir.yml files.
3. The new "--safe" flag on the "sync" command make sure there are no local commit or change before performing the actual synchronization.

This is the initial implementation of these features, and we are starting to use them internally.
Even though I expect some changes to be done based on our usage, a first review of the general approach would be nice, so I can take your feedback into consideration as soon as possible.

The main point I still need to address is how to enable the safe mode by default for a user or a project.

### DIFF
--- a/pkg/vendir/cmd/baseline.go
+++ b/pkg/vendir/cmd/baseline.go
@@ -1,3 +1,6 @@
+// Copyright 2024 The Carvel Authors.
+// SPDX-License-Identifier: Apache-2.0
+
 package cmd
 
 import (
@@ -14,7 +17,7 @@ import (
 	ctlcache "carvel.dev/vendir/pkg/vendir/fetch/cache"
 )
 
-func NewBaselineOptions(ui ui.UI) *BaselineOptions {
+func NewBaselineOptions(ui ui.UI) *BaselineOptions { //nolint:revive
 	return &BaselineOptions{ui: ui}
 }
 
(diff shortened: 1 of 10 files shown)


cdevienne:
Hi @rohitagg2020 

I did a few changes to make the linter happier, but if I take everything in consideration I feel the code gets less readable.
For example it wants constants for 0, 1 or 2, where it is clearly not a good idea.
The cyclomatic settings are pretty restrictive too, and the line max length at 80 feels a little short in some cases.

The thing is that I realized that most of the vendir code does not respect these settings: so, should I strictly comply to what the linter says ? or would it make sense to change 'revive' settings ?

pkg/vendir/fetch/hg/status.go
```diff
@@ -0,0 +1,93 @@
...
+		Ref: ctlstatus.CompleteReference{
+			SHA: sha,
+		},
+	}
+	if tags != emptyString {
+		status.Ref.Tags = strings.Split(tags, " ")
+		status.Ref.Tags = slices.DeleteFunc(
+			status.Ref.Tags, func(t string) bool { return t == "tip" })
+	}
+	if branch != emptyString {
```
joaopapereira, line 62: Can this piece of code not be just 
`status.Ref.Others = append(status.Ref.Others, branch, topic, bookmarks)`?
Because if it is an empty string it will append "" which should not be a problem

pkg/vendir/status/status.go
```diff
@@ -0,0 +1,156 @@
...
+	Others []string
+}
+
+type Status struct {
+	DirectoryPath     string
+	ContentPath       string
+	TargetRef         string
+	Ref               CompleteReference
+	UncommitedChanges []string
+	LocalCsets        []string
```
joaopapereira, line 27: With this name I am not sure what will be present in this variable.

pkg/vendir/cmd/baseline.go
```diff
@@ -0,0 +1,241 @@
...
+	}
+
+	statusMap, err := fullStatus(conf, syncOpts, existingLockConfig, o.ui)
+	if err != nil {
+		return err
+	}
+
+	newRefs := make(map[string]string)
+
+	for _, status := range statusMap {
```
joaopapereira, line 101: If you have uncommitted changes in a particular directory I think you should not allow rebasing because you are setting the state to something that cannot be reproducible.

pkg/vendir/cmd/baseline.go
```diff
@@ -0,0 +1,241 @@
...
+	if err := enc.Encode(doc); err != nil {
+		_ = f.Close()
+
+		return err
+	}
+
+	return f.Close()
+}
+
+func updateRefs(fname string, newRefs map[string]string) error {
```
joaopapereira, line 182: feels like we should use the same mechanism here to generate the lock and config file that we use in the sync commands. This "snipping" changes in the configuration can make future changes more complicated.

joaopapereira (changes requested):
Here are my initials impressions:
- The status looks pretty straightforward, despite the fact that I am not a fan of the tabular output I am not opposed.
- I would prefer the baseline and status to be in different PRs because they are quite different features, despite the fact that they reuse some internal functions
- In my opinion the baseline command should not be successful if there are uncommitted changes in a folder because we should not save none reproducible behaviors.
- Looks like we are not updating the lock file when we do the baseline command, that does not sound ok.
- in terms of the linter I am not sure why it is so mad lately but I am ok with not addressing the issues we see in the status (for more context here, we added the linter way later in the project and decided to fix the linter issues as we did changes in the code instead of doing a full sweep of the full code trying to fix everything)

Let me know if you have any question about my review. Also we need some tests to make sure this is working in the future if there is any change to this code.

PR #439, opened 2026-01-27
Testing : Fix semver comparision issue

sameerforge:


joaopapereira (changes requested):
hmm, This feel more like an LT problem for the Semver library more than this function in particular. We should open PR to fix that library.
Also this is cause breaking backward compatibility because this part of the code as not been changed in a very long time

sameerforge:
> hmm, This feel more like an LT problem for the Semver library more than this function in particular. We should open PR to fix that library. Also this is cause breaking backward compatibility because this part of the code as not been changed in a very long time

I have created a PR to fix the semver library for this issue. https://github.com/carvel-dev/semver/pull/5. cLosing this PR.

PR #453, opened 2026-06-24
perf(git): optimize fetch for repos with many tags and branches

zendesk-slowery23:
## Problem

`vendir sync` is very slow for git repositories with large numbers of branches and tags. The root cause is that the git fetcher unconditionally:
1. Set `remote.origin.tagOpt = --tags`, causing all tag objects to be downloaded on every fetch
2. Did not pass a refspec to `git fetch`, causing all remote refs to be negotiated

On a repo with thousands of branches and tags, this turns every sync into a multi-megabyte download regardless of what ref you actually need.

## Solution

When a fixed named ref is specified (tag or branch, not a SHA), use a targeted single-refspec fetch:

```
git fetch origin <ref> --no-tags [--depth N]
```

Then check out `FETCH_HEAD` (which git populates from a single-refspec fetch) instead of the ref name.

`--tags` is kept only where it's actually needed:
- `refSelection` (semver matching must enumerate all tags)
- Plain SHA refs with no `OriginalRef` (SHAs can't be fetched by name)

### `--locked` path

`--locked` substitutes the locked SHA for the original ref before reaching the fetcher, which previously forced a full `--tags` fetch every time. This PR preserves the original ref in a new unexported `OriginalRef` field on `DirectoryContentsGit` (set by `Lock()`, not serialized to YAML). The fetcher uses it for the same targeted fetch path, then verifies the resulting commit SHA matches the lock — catching force-pushed tags.

## Performance

Measured against an internal repo with thousands of branches and hundreds of tags:

| Command | Before | After | Improvement |
|---|---|---|---|
| `vendir sync` | 43s | 11s | **74% faster** |
| `vendir sync --locked` | 1m 26s | 9s | **89% faster** |

## Changes

- `pkg/vendir/config/directory.go` — add `OriginalRef string` (unexported from YAML) to `DirectoryContentsGit`; set it in `Lock()` before overwriting `Ref` with the SHA
- `pkg/vendir/fetch/git/git.go` — targeted refspec fetch for named refs and locked-SHA-with-OriginalRef; conditional `--tags`/`--no-tags`; post-checkout SHA verification in locked mode; `isHexSHA()` helper
- `pkg/vendir/fetch/git/git_test.go` — unit tests for each fetch path (named tag, `origin/` branch, locked SHA+OriginalRef, plain SHA, depth ordering)
- `test/e2e/git_test.go` — `TestGitFetchOptimizations`: verifies targeted fetch line in `--json` output, locked sync uses OriginalRef, tampered lock SHA is detected

## Testing

```bash
./hack/build.sh
go test ./pkg/vendir/fetch/git/... -v
VENDIR_BINARY_PATH=./vendir go test ./test/e2e/... -v -run TestGitFetchOptimizations
```

🤖 Generated with [Claude Code](https://claude.ai/claude-code)

### DIFF
--- a/pkg/vendir/fetch/git/git.go
+++ b/pkg/vendir/fetch/git/git.go
@@ -166,30 +166,81 @@ func (t *Git) fetch(dstPath string, tempArea ctlfetch.TempArea, bundle string) e
 		}
 	}
 
-	// A targeted refspec fetch (named ref, or locked SHA with a known
-	// OriginalRef) doesn't need all tags. RefSelection needs them for semver
-	// matching, and a bare SHA with no OriginalRef can't be targeted at all.
-	tagOpt := "--tags"
-	if t.canTargetFetch() {
-		tagOpt = noTagsFlag
-	}
-	argss = append(argss, []string{"config", "remote.origin.tagOpt", tagOpt})
-
 	if bundle != "" {
 		argss = append(argss, []string{"bundle", "unbundle", bundle})
 	}
 
-	// When fetching a single named ref, git places it in FETCH_HEAD only —
-	// no local ref is created. Track this to use FETCH_HEAD for checkout.
-	fetchArgs, useFetchHead := t.targetedFetchArgs()
-	if t.opts.Depth > 0 {
-		fetchArgs = append(fetchArgs, "--depth", strconv.Itoa(t.opts.Depth))
-	}
-	argss = append(argss, fetchArgs)
-
 	err = t.cmdRunner.RunMultiple(argss, env, dstPath)
 	if err != nil {
 		return err
+	}
+
+	// Resolve an ambiguous ref (which may be either a branch or a tag) before
+	// constructing the refspec. This lets us preserve tags without fetching
+	// every tag in the repository.
+	refType := remoteRefUnknown
+	fetchRef := t.opts.Ref
+	if t.isLockedSHAWithNamedOriginal() {
+		fetchRef = t.opts.OriginalRef
+	}
+	if t.canTargetFetch() {
+		var err error
+		refType, err = t.remoteRefType(dstPath, fetchRef)
+		if err != nil {
+			return err
+		}
+	}
+
+	// RefSelection needs all tags for semver matching, and a bare SHA has no
+	// named ref to target. A targeted tag fetch explicitly creates just that
+	// tag; a targeted branch fetch uses Git's default auto-follow behavior,
+	// which fetches only tags reachable from the branch tip.
+	if !t.canTargetFetch() || refType == remoteRefUnknown ||
+		refType == remoteRefTag {
+		tagOpt := "--tags"
+		if t.canTargetFetch() {
+			tagOpt = noTagsFlag
+		}
+		_, _, err = t.cmdRunner.Run(
+			[]string{"config", "remote.origin.tagOpt", tagOpt}, nil, dstPath)
+		if err != nil {
+			return err
+		}
+	}
+
+	// When fetching a single named ref, track the result in FETCH_HEAD for
+	// checkout. Branches/tags also get local refs to support bundle caching.
+	fetchArgs, useFetchHead := t.targetedFetchArgs(fetchRef, refType)
+	if t.opts.Depth > 0 {
+		fetchArgs = append(fetchArgs, "--depth", strconv.Itoa(t.opts.Depth))
+	}
+
+	_, _, err = t.cmdRunner.Run(fetchArgs, env, dstPath)
+	lockedFetchByOriginal := t.isLockedSHAWithNamedOriginal() &&
+		fetchRef != t.opts.Ref
+	if err != nil && lockedFetchByOriginal {
+		// The original ref may have been deleted or renamed since the config
+		// was locked. Fetching the locked object directly retains locked-sync
+		// behavior without requiring the old name to remain available.
+		fallbackArgs := []string{"fetch", "origin", t.opts.Ref, noTagsFlag}
+		if t.opts.Depth > 0 {
+			fallbackArgs = append(
+				fallbackArgs, "--depth", strconv.Itoa(t.opts.Depth),
+			)
+		}
+		_, _, err = t.cmdRunner.Run(fallbackArgs, env, dstPath)
+	}
+	if err != nil {
+		return err
+	}
+
+	if useFetchHead {
+		_, _, err = t.cmdRunner.Run([]string{
+			"update-ref", "refs/vendir/fetched", "FETCH_HEAD",
+		}, nil, dstPath)
+		if err != nil {
+			return err
+		}
 	}
 
 	ref, err := t.resolveRef(dstPath)
@@ -241,9 +292,13 @@ func (t *Git) fetch(dstPath string, tempArea ctlfetch.TempArea, bundle string) e
 }
 
 const (
-	noTagsFlag      = "--no-tags"
-	fetchHeadCommit = "FETCH_HEAD^{commit}"
-	originPrefix    = "origin/"
+	noTagsFlag         = "--no-tags"
+	fetchHeadCommit    = "FETCH_HEAD^{commit}"
+	originName         = "origin"
+	originPrefix       = originName + "/"
+	remoteHeadsPrefix  = "refs/heads/"
+	remoteTagsPrefix   = "refs/tags/"
+	minRemoteRefFields = 2
 )
 
 // isLockedSHAWithNamedOriginal reports whether Ref is a locked SHA with a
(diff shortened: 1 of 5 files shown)


pkg/vendir/fetch/http/sync.go
```diff
@@ -195,8 +195,7 @@ func (t *Sync) addAuth(req *http.Request) error {
 
 	// Basic auth — password is optional, defaults to empty string
 	if hasUser {
-		password := string(secret.Data[
-			ctlconf.SecretK8sCorev1BasicAuthPasswordKey])
+		password := string(secret.Data[ctlconf.SecretK8sCorev1BasicAuthPasswordKey])
```
zendesk-slowery23: result of `go fmt` from `hack/build.sh`

pkg/vendir/fetch/git/git.go
```diff
@@ -166,17 +166,40 @@ func (t *Git) fetch(dstPath string, tempArea ctlfetch.TempArea, bundle string) e
...
 			fetchArgs = append(fetchArgs, t.opts.Ref[7:])
+		case len(t.opts.Ref) > 0 && !isHexSHA(t.opts.Ref):
+			// fetch only the named tag or branch; SHAs must use a full fetch
+			fetchArgs = append(fetchArgs, t.opts.Ref, "--no-tags")
+			useFetchHead = true
+		case isHexSHA(t.opts.Ref) && len(t.opts.OriginalRef) > 0 && !isHexSHA(t.opts.OriginalRef):
+			// locked mode: fetch by the original named ref so we avoid a full fetch
+			fetchArgs = append(fetchArgs, t.opts.OriginalRef, "--no-tags")
+			useFetchHead = true
 		}
```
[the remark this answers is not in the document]
zendesk-slowery23: Good catch. Fixed in the follow-up commit — `OriginalRef` now has `origin/` stripped before being used as a fetch refspec: `strings.TrimPrefix(t.opts.OriginalRef, "origin/")`. Also added a unit test case specifically for this path (`Locked SHA with origin/ OriginalRef strips prefix before fetch`).

pkg/vendir/fetch/git/git.go
```diff
@@ -193,6 +216,9 @@ func (t *Git) fetch(dstPath string, tempArea ctlfetch.TempArea, bundle string) e
 	if err != nil {
 		return err
 	}
+	if useFetchHead {
+		ref = "FETCH_HEAD"
+	}
```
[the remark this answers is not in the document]
zendesk-slowery23: Good point. Fixed in the follow-up commit. The locked-mode path now resolves `FETCH_HEAD^{commit}` and compares it to the locked SHA *before* checkout, then checks out the locked SHA directly. That way a force-pushed tag is still caught, but mutable refs like `origin/main` stay pinned to the exact locked commit rather than the current branch tip.

pkg/vendir/fetch/git/git.go
```diff
@@ -245,6 +283,21 @@ func (t *Git) tags(dstPath string) ([]string, error) {
...
+	if len(s) < 7 || len(s) > 40 {
+		return false
+	}
+	for _, c := range s {
+		if !((c >= '0' && c <= '9') || (c >= 'a' && c <= 'f')) {
+			return false
+		}
+	}
+	return true
+}
```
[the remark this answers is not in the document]
zendesk-slowery23: Fixed — `isHexSHA` now accepts `A-F` in addition to `a-f`, matching git's case-insensitive treatment of object IDs.

test/e2e/git_test.go
```diff
@@ -235,6 +235,131 @@ directories:
...
+
+		var vendirOutput VendirOutput
+		require.NoError(t, json.NewDecoder(&stdout).Decode(&vendirOutput))
+
+		var foundTargetedFetch bool
+		for _, l := range vendirOutput.Lines {
+			if strings.Contains(l, "fetch origin signed-trusted-tag --no-tags") {
+				foundTargetedFetch = true
+			}
+			assert.NotContains(t, l, "--> git fetch origin\n", "expected no bare full fetch for named tag ref")
```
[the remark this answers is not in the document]
zendesk-slowery23: You're right — the `\n` in the needle meant `NotContains` would never trigger since `Lines` entries don't include trailing newlines. Replaced with a positive `foundBareFetch` flag that checks for `"--> git fetch origin"` without `"signed-trusted-tag"` in the same line, then asserts it was never seen.

pkg/vendir/fetch/git/git_test.go
```diff
@@ -65,10 +66,159 @@ func TestGit_Retrieve(t *testing.T) {
...
+		assert.Contains(t, checkoutArgs, "main", "expected branch name checkout for origin/ ref")
+		assert.NotContains(t, checkoutArgs, "FETCH_HEAD", "expected direct checkout for origin/ ref, not FETCH_HEAD")
+	})
+
+	t.Run("Locked SHA with OriginalRef uses OriginalRef for targeted fetch", func(t *testing.T) {
+		lockedSHA := "a1b2c3d4e5f6a1b2c3d4e5f6a1b2c3d4e5f6a1b2"
+		runner := &cmdRunnerLocal{
+			commandsToRun: [][]string{},
+			// rev-parse HEAD must return the locked SHA for the post-checkout verification to pass
+			responses: map[string]string{"rev-parse HEAD": lockedSHA},
```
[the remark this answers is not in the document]
zendesk-slowery23: Updated to stub `rev-parse FETCH_HEAD^{commit}` instead, matching the revised verification approach (checking the fetched commit before checkout rather than HEAD after).

pkg/vendir/fetch/git/git_test.go
```diff
@@ -65,10 +66,159 @@ func TestGit_Retrieve(t *testing.T) {
...
+		assert.Contains(t, fetchArgs, "--no-tags", "expected --no-tags for locked SHA with OriginalRef")
+
+		tagOptArgs := findConfigArgs(runner.commandsToRun, "remote.origin.tagOpt")
+		require.NotNil(t, tagOptArgs)
+		assert.Equal(t, "--no-tags", tagOptArgs[len(tagOptArgs)-1], "expected --no-tags tagOpt for locked SHA with OriginalRef")
+
+		checkoutArgs := findCheckoutArgs(runner.commandsToRun)
+		require.NotNil(t, checkoutArgs)
+		assert.Contains(t, checkoutArgs, "FETCH_HEAD", "expected FETCH_HEAD checkout in locked SHA mode")
+
```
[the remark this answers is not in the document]
zendesk-slowery23: Fixed — the locked-mode test now asserts that checkout uses the locked SHA directly, not `FETCH_HEAD`, and that `rev-parse FETCH_HEAD^{commit}` is called for pre-checkout verification.

zendesk-slowery23:
Are any of the maintainers available to approve the workflows for my PR and/or to discuss it?  We use vendir pretty heavily and this performance improvement will have a big boost in our engineer's productivity.

zendesk-slowery23:
Nudge on this.  Anybody still maintaining vendir?

joaopapereira (approved):
Sorry about the delay, Lgtm

joaopapereira:
Do you mind reading this or so that we have the tests back? Thanks

zendesk-slowery23:
Pushed a fix for the fetch bug that was causing `Test GH` and `lint` to fail — a mismatched FETCH_HEAD/checkout case for `origin/<tag>` refs, plus the lint issues Copilot/CI flagged.

I ran both `go test ./pkg/...` and `golangci-lint run ./...` (pinned to v2.4.0, matching CI's config) locally and both pass on the changed files. I also manually reproduced the exact e2e scenarios that were failing (tag ref, `origin/`-prefixed tag ref, branch ref, `--locked` sync, and moved-tag detection) against a real repo and confirmed they now work correctly.

That said, I can't confirm the actual GitHub Actions checks will pass since they're still pending workflow approval on this PR — happy to fix anything else that surfaces once they run. Thanks for flagging this!

joaopapereira:
I just let them run and they are failing do you mind taking a look?

zendesk-slowery23:
Thanks for letting me know. I've tested the changes locally, including the previously failing Git/cache/locked E2E cases and lint, and I think they should address the checks that are awaiting validation. I pushed the fixes in `d29bc66b`; the updated workflows are currently awaiting approval to run.

zendesk-slowery23:
Looks like everything is green now 🎉 .  Any reason to hold off on merging/releasing?

pkg/vendir/fetch/git/git.go
```diff
@@ -166,39 +166,111 @@ func (t *Git) fetch(dstPath string, tempArea ctlfetch.TempArea, bundle string) e
...
 
-	if t.opts.Verification != nil {
-		verification := Verification{
-			dstPath, *t.opts.Verification, t.refFetcher, t.infoLog,
+	// In locked mode we fetch by the original named ref for speed, but must
+	// check out the exact locked SHA. Verify FETCH_HEAD resolves to that SHA
+	// before checkout so a force-pushed tag is caught early.
+	if useFetchHead && isHexSHA(t.opts.Ref) {
+		rp := []string{"rev-parse", fetchHeadCommit}
+		out, _, runErr := t.cmdRunner.Run(rp, nil, dstPath)
```
joaopapereira, line 256: Not passing env so we lose all the env variables that we might have set

pkg/vendir/fetch/git/git.go
```diff
@@ -219,13 +291,107 @@ func (t *Git) fetch(dstPath string, tempArea ctlfetch.TempArea, bundle string) e
...
+func (t *Git) remoteRefType(dstPath, ref string) (remoteRefType, error) {
+	normalizedRef := strings.TrimPrefix(ref, originPrefix)
+	switch {
+	case strings.HasPrefix(normalizedRef, remoteHeadsPrefix):
+		return remoteRefBranch, nil
+	case strings.HasPrefix(normalizedRef, remoteTagsPrefix):
+		return remoteRefTag, nil
+	}
+
+	out, _, err := t.cmdRunner.Run(
```
joaopapereira, line 367: Not passing env so we lose all the env variables that we might have set

pkg/vendir/fetch/git/git.go
```diff
@@ -166,39 +166,111 @@ func (t *Git) fetch(dstPath string, tempArea ctlfetch.TempArea, bundle string) e
...
 		return err
 	}
 
-	if t.opts.Verification != nil {
-		verification := Verification{
-			dstPath, *t.opts.Verification, t.refFetcher, t.infoLog,
+	// In locked mode we fetch by the original named ref for speed, but must
+	// check out the exact locked SHA. Verify FETCH_HEAD resolves to that SHA
+	// before checkout so a force-pushed tag is caught early.
+	if useFetchHead && isHexSHA(t.opts.Ref) {
```
joaopapereira, line 254: the "fetched SHA must equal locked SHA" hard-check runs even when `OriginalRef` is a branch (not a tag). Since branches move, running `vendir sync --locked` after the branch has advanced normally will now hard-fail with a "tag may have been moved" error, when the locked commit is still perfectly valid and reachable.

pkg/vendir/fetch/git/git.go
```diff
@@ -166,39 +166,111 @@ func (t *Git) fetch(dstPath string, tempArea ctlfetch.TempArea, bundle string) e
...
+	}
+
+	_, _, err = t.cmdRunner.Run(fetchArgs, env, dstPath)
+	lockedFetchByOriginal := t.isLockedSHAWithNamedOriginal() &&
+		fetchRef != t.opts.Ref
+	if err != nil && lockedFetchByOriginal {
+		// The original ref may have been deleted or renamed since the config
+		// was locked. Fetching the locked object directly retains locked-sync
+		// behavior without requiring the old name to remain available.
+		fallbackArgs := []string{"fetch", "origin", t.opts.Ref, noTagsFlag}
```
joaopapereira, line 225: when the named-ref targeted fetch fails, it falls back to `git fetch origin <sha> --no-tags`. Many git servers disallow fetching an arbitrary non-tip SHA (uploadpack.allowReachableSHA1InWant not enabled). Pre-PR, a full mirror fetch would have pulled the SHA in via any other ref containing it; this fallback has no further recovery path.

pkg/vendir/fetch/git/git.go
```diff
@@ -219,13 +291,107 @@ func (t *Git) fetch(dstPath string, tempArea ctlfetch.TempArea, bundle string) e
...
+	}
+
+	branchRef := remoteHeadsPrefix + normalizedRef
+	tagRef := remoteTagsPrefix + normalizedRef
+	for _, line := range strings.Split(out, "\n") {
+		fields := strings.Fields(line)
+		if len(fields) < minRemoteRefFields {
+			continue
+		}
+		switch fields[1] {
```
joaopapereira, line 380: remoteRefType() returns on the first ls-remote match; since output is alphabetically sorted,` refs/heads/<name>` sorts before `refs/tags/<name>`, so a name that exists as both is always classified as a branch. This is the opposite of git's own precedence (tags win per gitrevisions(7)), so a repo with both `v2` branch and `v2` tag will now checkout a different commit than before.

pkg/vendir/fetch/git/git.go
```diff
@@ -248,6 +414,37 @@ func (t *Git) tags(dstPath string) ([]string, error) {
...
 
+const (
+	minAbbreviatedSHALen = 7
+	fullSHALen           = 40
+)
+
+// isHexSHA reports whether s looks like a full or abbreviated git commit SHA
+// (7–40 hex characters, case-insensitive). SHAs cannot be fetched by refspec
+// name and require a full fetch to be resolved.
+func isHexSHA(s string) bool {
```
joaopapereira, line 425: This function can produce false positives because it is possible to have words or mix of words with numbers that look HEX but are just normal branch names ex `cafebabe` or something.

pkg/vendir/fetch/git/git.go
joaopapereira, line 84: Despite pkg/vendir/fetch/git/sync.go:55 was not modified by this PR, the URL-only bundle cache key predates it. But git.go's new targeted-fetch behavior means a sync now typically populates the local bundle with only one ref's objects, instead of the full mirror the old unconditional full-fetch produced. Since the cache key doesn't include the ref, a vendir.yml with two directories: entries on the same URL but different refs can now overwrite each other's cached bundle each run, a caching regression this PR introduces by changing what goes into the bundle, even though it doesn't touch the cache-key code itself.

pkg/vendir/fetch/git/git.go
```diff
@@ -166,39 +166,111 @@ func (t *Git) fetch(dstPath string, tempArea ctlfetch.TempArea, bundle string) e
...
+		if err != nil {
+			return err
+		}
+	}
+
+	// RefSelection needs all tags for semver matching, and a bare SHA has no
+	// named ref to target. A targeted tag fetch explicitly creates just that
+	// tag; a targeted branch fetch uses Git's default auto-follow behavior,
+	// which fetches only tags reachable from the branch tip.
+	if !t.canTargetFetch() || refType == remoteRefUnknown ||
```
joaopapereira, line 198: targeted branch fetches leave tagOpt at git's default (auto-follow tags on newly transferred objects only) instead of the old unconditional `--tags`, so `git describe --tags HEAD` in Retrieve() can silently produce worse/empty tags: data in `vendir.lock.yml`, especially with depth: 1 or a warm bundle cache.

joaopapereira: Ex: vendir.yml has ref: main. Someone tagged an older commit on main (e.g., v1.2.0 on a commit from last week) as a release marker. On a sync where the bundle cache already has most of main's history, git only transfers the handful of new commits, v1.2.0's commit isn't among them, so auto-follow skips that tag. Later, Retrieve() calls git describe --tags HEAD to populate tags: in vendir.lock.yml, and it comes back empty or points to some other/older tag it does know about. Silently less accurate lock-file metadata than before the PR, with no error raised.

joaopapereira (changes requested):
Sorry for the late review, I think the PR is going in the right direction but there are some points that we will have to address in order to be able to get this merged

PR #457, opened 2026-08-04
Suppress "Succeeded" line after vendir completion output

### LINKED ISSUE #405: `vendir completion` outputs superflous "Succeeded"
**What steps did you take:**
[A clear and concise description steps that can be used to reproduce the problem.]

`vendir completion bash` outputs trailing "Succeeded" line, causing an error when sourced.

**What happened:**

```shell
vendir completion bash | tail -n 1
```

**What did you expect:**
[A description of what was expected]

No trailing "Succeeded" line.

[Additional information that will assist in solving the issue.]

**Environment:**

- vendir version (execute `vendir --version`):

---
Vote on this request

pujitha24:
Motivation:
`vendir completion bash` (and zsh/fish/powershell) prints a shell
script to stdout that is meant to be sourced or eval'd directly, e.g.
`source <(vendir completion bash)`. main() unconditionally printed a
trailing "Succeeded" line after every command, including completion,
so the sourced script ended with a bare `Succeeded` token, which the
shell then tried to execute as a command.

Approach:
main() now captures the actually-executed *cobra.Command via
command.ExecuteC() (Execute() is just ExecuteC() with the command
discarded, so this is behavior-preserving for every other command)
and skips the "Succeeded" line when the executed command is cobra's
built-in "completion" command or one of its bash/zsh/fish/powershell
subcommands. vendir does not define its own completion command, so
this only affects cobra's default completion tree.

An initial attempt mirrored the fix already applied in carvel-dev/kapp
(cmd/kapp/kapp.go), which guards the same print with
cobrautil.IsCobraManagedCommand(os.Args). Building and testing that
approach against this issue showed it does not actually suppress the
line for `completion bash`: IsCobraManagedCommand only special-cases
"help", "__complete" and "__completeNoDesc", not "completion" itself,
so it does not resolve this report.

Validation:
- go build ./...
- ./hack/test.sh (repo's documented unit test script) - all packages
  pass; this code path has no existing unit tests.
- ./hack/build.sh (repo's documented build script: go mod
  vendor/tidy, go fmt, real build, `./vendir version`, compile-only
  `go test --exec=echo ./...`) - all steps pass.
- Manually built the binary before and after the change and compared:
  `vendir completion bash|zsh|fish|powershell | tail -n 3` no longer
  ends with "Succeeded" (it did before), while `vendir version` and
  an error path (`vendir sync -f /nonexistent.yml`, exit 1) are
  unaffected and still print/omit "Succeeded" as before.
- `bash -c 'source <(vendir completion bash) && echo SOURCE_OK'`
  succeeds after the fix, reproducing and resolving the exact failure
  mode from the report (it fails without the fix, since the shell
  tries to run the stray `Succeeded` token as a command).

User-visible behavior for every non-completion command is unchanged.

Report: https://github.com/carvel-dev/vendir/issues/405
Signed-off-by: Pujitha Paladugu <10557236+pujitha24@users.noreply.github.com>

Fixes #405

### DIFF
--- a/cmd/vendir/vendir.go
+++ b/cmd/vendir/vendir.go
@@ -13,6 +13,7 @@ import (
 	"carvel.dev/vendir/pkg/vendir/cmd"
 	uierrs "github.com/cppforlife/go-cli-ui/errors"
 	"github.com/cppforlife/go-cli-ui/ui"
+	"github.com/spf13/cobra"
 )
 
 func main() {
@@ -28,11 +29,20 @@ func main() {
 
 	command := cmd.NewDefaultVendirCmd(confUI)
 
-	err := command.Execute()
+	executedCmd, err := command.ExecuteC()
 	if err != nil {
 		confUI.ErrorLinef("vendir: Error: %v", uierrs.NewMultiLineError(err))
 		os.Exit(1)
 	}
 
-	confUI.PrintLinef("Succeeded")
+	// The "completion" command (and its shell subcommands) prints a script
+	// that is meant to be sourced/eval'd directly, so it must not be
+	// followed by any extra output.
+	if !isCompletionCmd(executedCmd) {
+		confUI.PrintLinef("Succeeded")
+	}
+}
+
+func isCompletionCmd(c *cobra.Command) bool {
+	return c.Name() == "completion" || (c.HasParent() && c.Parent().Name() == "completion")
 }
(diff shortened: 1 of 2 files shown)


pujitha24:
Just checking in on this one - it's still rebased and passing checks. Happy to make any changes if something would help move it along.

joaopapereira (approved):
Lgtm

PR #459, opened 2026-08-14
fix: leave already up to date files untouched on sync

Allsimon:
vendir sync replaced the whole final location on every run: it deleted the directory and renamed the staging directory over it. 
Every file came back with a fresh mtime and inode even when nothing had changed, which wakes up file watchers, build caches and IDEs for no reason.

Reconcile the staging directory against the final location instead. Regular files whose size and sha256 match, and symlinks pointing at the same target, are left alone; only changed and new entries are moved over, stale ones are deleted, and permissions are aligned with a chmod, which does not touch mtimes.

Manually managed contents are no longer staged at all. They used to be renamed out to the staging directory and back, which left their parent directory looking modified. The reconcile is told to preserve them instead, so vendir does not touch them at all.

### DIFF
--- a/pkg/vendir/cmd/sync.go
+++ b/pkg/vendir/cmd/sync.go
@@ -28,9 +28,10 @@ type SyncOptions struct {
 	Files    []string
 	LockFile string
 
-	Directories []string
-	Locked      bool
-	Lazy        bool
+	Directories   []string
+	Locked        bool
+	Lazy          bool
+	MergeDiffOnly bool
 
 	Chdir                       string
 	AllowAllSymlinkDestinations bool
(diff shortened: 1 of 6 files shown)


joaopapereira (commented):
I marked has a comment when I wanted to mark it as a request for changes 😢

joaopapereira (changes requested):
Thanks for the PR. The reconcile approach is well-built and the time-stability win is real. My main hesitation is performance: every unchanged file now gets fully read and sha256'd on both the staging and final sides on every sync, where before it was a single rename. On large vendored trees (big helm charts, image filesystems) that turns a no-op resync into "read every byte, twice."

Rather than block on that, I'd like this behavior to be opt-in behind a flag so the default keeps today's fast path. Something like `--merge-diff-only` (name open to bikeshedding), wired the same way as `--lazy`: a bool on SyncOptions, plumbed through SyncOpts into Directory.Sync.

One thing to watch when you implement it: the reconcile and the "manual contents are no longer staged" change are coupled. If the flag is off and you fall back to os.Rename swapping the whole staging dir over the final location, the in-place manual contents get deleted (staging no longer contains them). So the flag needs to gate both the manual-contents branch and the final Replace/PartialRepace, keeping the legacy staging+rename path alive for the off case.

A couple of decisions worth your call:
- Alternative worth considering: instead of gating the whole materialization strategy (which means maintaining two code paths forever), gate just the comparison: a size+mtime fast-path in sameEntry that skips hashing when both match. That kills the perf cost for the common no-op resync without a legacy path, at the cost of a weaker guarantee than content hashing. Could be its own flag (--verify=size-mtime|content) if you want to keep content-hash available.

Could you add the flag (or the fast-path) before we merge? Also please make sure the manual-contents + permissions behavior has end-to-end test coverage, since the new unit tests stop at the reconcileDir boundary. This should be a opt-in behavior because we need to keep backwards compatibility.

